PR #3, opened 2017-06-20
I333

littledice:
return coverage set orders with one-indexing

### DIFF
--- a/src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqTouchstoneRepository.kt
+++ b/src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqTouchstoneRepository.kt
@@ -133,7 +133,7 @@ class JooqTouchstoneRepository(private val scenarioRepository: () -> JooqScenari
     fun mapCoverageRow(record: Record, scenarioDescriptionId: String) = CoverageRow(
             scenarioDescriptionId,
             record[COVERAGE_SET.ID],
-            record[SCENARIO_COVERAGE_SET.ORDER],
+            record[SCENARIO_COVERAGE_SET.ORDER] + 1,
             record[COVERAGE_SET.NAME],
             record[COVERAGE_SET.VACCINE],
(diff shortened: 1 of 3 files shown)


src/blackboxTests/src/test/kotlin/org/vaccineimpact/api/blackboxTests/schemas/CSVSchema.kt
```diff
@@ -12,7 +12,7 @@ class CSVSchema(schemaFileName: String)
 {
     val columns = parseSchema(schemaFileName)
 
-    fun validate(csvAsString: String)
+    fun validate(csvAsString: String): Iterable<IndexedValue<Array<String>>>
```
MartinEden: Do you need this to be indexed? In `coverage sets are indexed from one` you invoke `csvRows.first()`, which I think you can do without having applied indexing.

src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqTouchstoneRepository.kt
```diff
@@ -133,7 +133,7 @@ class JooqTouchstoneRepository(private val scenarioRepository: () -> JooqScenari
     fun mapCoverageRow(record: Record, scenarioDescriptionId: String) = CoverageRow(
             scenarioDescriptionId,
             record[COVERAGE_SET.ID],
-            record[SCENARIO_COVERAGE_SET.ORDER],
+            record[SCENARIO_COVERAGE_SET.ORDER] + 1,
```
MartinEden: It might be nice to have a comment here, explaining why we are adding one to this value. Although, now I think about that, did I even explain that to you? The issue here is that CSV file is going to be consumed by academics, who are in general not programmers, and not familiar with zero indexing. This is just to make things a bit more human friendly.

MartinEden:
I've added a few very minor points of review, but I'd accept this code as is. Feel free to fix the niggles or just merged the PR. Nice one :)

PR #4, opened 2017-06-20
Add CLI function to add user to all modelling groups

MartinEden:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-342

### DIFF
--- a/src/databaseInterface/src/main/kotlin/org/vaccineimpact/api/db/JooqContext.kt
+++ b/src/databaseInterface/src/main/kotlin/org/vaccineimpact/api/db/JooqContext.kt
@@ -6,7 +6,7 @@ import org.jooq.impl.DSL
 import java.sql.Connection
 import java.sql.DriverManager
 
-class JooqContext(val dbName: String? = null) : AutoCloseable
+open class JooqContext(val dbName: String? = null) : AutoCloseable
 {
     private val conn = getConnection()
     val dsl = createDSL(conn)
--- a/src/userCLI/src/main/kotlin/org/vaccineimpact/api/security/AddUserToGroup.kt
+++ b/src/userCLI/src/main/kotlin/org/vaccineimpact/api/security/AddUserToGroup.kt
@@ -11,7 +11,7 @@ fun addToGroup(args: List<String>)
 
 data class AddUserToGroupOptions(
         val username: String,
-        val groups: GroupDefinition
+        val groups: Groups
 )
 {
     companion object
@@ -34,11 +34,11 @@ data class AddUserToGroupOptions(
                     {
                         throw ActionException("Cannot combine the 'ALL' groups option with a list of groups")
                     }
-                    GroupDefinition.AllGroups()
+                    Groups.AllGroups()
                 }
                 else
                 {
-                    GroupDefinition.GroupList(args.drop(1))
+                    Groups.GroupList(args.drop(1))
                 }
                 return AddUserToGroupOptions(username, groups)
             }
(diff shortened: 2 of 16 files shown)


src/userCLI/src/main/kotlin/org/vaccineimpact/api/security/AddUserToGroup.kt
```diff
@@ -0,0 +1,71 @@
...
+import org.vaccineimpact.api.models.permissions.ReifiedRole
+import kotlin.system.exitProcess
+
+sealed class GroupDefinition
+{
+    abstract fun getGroups(db: JooqContext): Iterable<String>
+
+    class GroupList(val groups: List<String>): GroupDefinition()
+    {
+        override fun getGroups(db: JooqContext) = groups
```
littledice: this method doesn't actually use the argument passed in; seems like this whole sealed class is overkill - all you really need is one method, getAllGroups

src/userCLI/src/main/kotlin/org/vaccineimpact/api/security/AddUserToGroup.kt
```diff
@@ -0,0 +1,71 @@
...
+                    .from(MODELLING_GROUP)
+                    .fetch()
+                    .map { it[MODELLING_GROUP.ID] }
+        }
+    }
+}
+
+class AddUserToGroupOptions(
+        val username: String,
+        val groups: GroupDefinition
```
littledice: so here I would just make this a list of strings (group names), line 48 you call the 'getAllGroups' function, line 52 you just take the list of command line args

src/databaseInterface/src/main/kotlin/org/vaccineimpact/api/db/JooqContext.kt
```diff
@@ -6,7 +6,7 @@ import org.jooq.impl.DSL
 import java.sql.Connection
 import java.sql.DriverManager
 
-class JooqContext(val dbName: String? = null) : AutoCloseable
+open class JooqContext(val dbName: String? = null) : AutoCloseable
```
MartinEden: This is so we can mock it in one of the tests

MartinEden:
I've added tests - thank you for the prod on that. I was avoiding it because it meant adding a new test project, but I think it's worth doing. I've tried to also rewrite and reorganise the code a bit to make it have a simpler flow.

As to the sealed class/group definition business: There's definitely a trade off on complexity here. I may well have gone too far. I'll just explain why I've done it this way, and if you still think it should change I will do so forthwith.

The thing I wanted to avoid was conflating parsing the arguments with accessing the database. The sealed class allows you to capture the _intent_ of the user (either a specific list of groups or just all the groups) without having to actually go to the database right there and then and pull out the full list of groups. This is most easily seen in the tests I've just written (thanks again!) as with the current structure I can test the parsing and the retrieval as two separate concerns.

What do you think? Still a step too far for such an otherwise simple program?

MartinEden:
Oh, as to the name `GroupDefinition`: I'd welcome a better name. It's trying to capture what groups the user wanted. Maybe `GroupOption`, instead?

littledice:
Nice one on the tests. 

Yeah, I like the separation of parsing and db access. When I suggested adding the db access functions directly to the AddUserToGroupOptions I was imagining:

    class AddUserToGroupOptions(
        val username: String,
        val getGroups: (db: JooqContext) -> Iterable<String>
    )

Anyway, up to you. If you want to keep the class wrapper around the functions, maybe just Groups is a little clearer than GroupDefinition?

MartinEden:
I think I'll keep it with the separate class, as its easier to test this way. Otherwise, I would have to check that the function returned by `parseArgs` was the expected one, which seems tricky.

I have renamed `GroupDefinition` to `Groups`. Okay with me to merge this if and when it passes the tests?

PR #6, opened 2017-06-22
I90

littledice:
First iteration of the /users/:username endpoint - returns user with roles for all logged in API users without checking permissions.

### DIFF
--- a/src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/UserController.kt
+++ b/src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/UserController.kt
@@ -15,7 +15,7 @@ class UserController(context: ControllerContext) : AbstractController(context)
     fun getUser(context: ActionContext): User
     {
         var userName = userName(context)
-        var user = repos.user().use { it.getUserByUsername(userName(context))  }
+        var user = repos.user().use { it.getUserByUsername(userName)  }
         if (user == null)
             throw UnknownObjectError(userName, "Username")
 
--- a/src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/UserController.kt
+++ b/src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/UserController.kt
@@ -1,0 +1,33 @@
+package org.vaccineimpact.api.app.controllers
+
+import org.vaccineimpact.api.app.ActionContext
+import org.vaccineimpact.api.app.controllers.endpoints.BasicEndpoint
+import org.vaccineimpact.api.app.controllers.endpoints.SecuredEndpoint
+import org.vaccineimpact.api.app.errors.UnknownObjectError
+import org.vaccineimpact.api.models.UserDto
+
+class UserController(context: ControllerContext) : AbstractController(context)
+{
+    override val urlComponent = "/users"
+    override val endpoints = listOf(
+          //  BasicEndpoint("/", this::getUsers),
+            SecuredEndpoint("/:username/", this::getUser, emptySet())
+    )
+
+    fun getUser(context: ActionContext): UserDto
+    {
+        var userName = userName(context)
+        var user = repos.user().use { it.getUserByUsername(userName(context))  }
+        if (user == null)
+            throw UnknownObjectError(userName, "Username")
+
+        return user
+    }
+//
+//    fun getUsers(context: ActionContext) : Result<Record>?
+//    {
+//        return repos.user().use { it.getAllUsers() }
+//    }
+
+    private fun userName(context: ActionContext): String = context.params(":username")
+}
--- a/src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/UserController.kt
+++ b/src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/UserController.kt
@@ -15,11 +15,8 @@ class UserController(context: ControllerContext) : AbstractController(context)
     fun getUser(context: ActionContext): User
     {
         var userName = userName(context)
-        var user = repos.user().use { it.getUserByUsername(userName)  }
-        if (user == null)
-            throw UnknownObjectError(userName, "Username")
 
-        return user
+        return repos.user().use { it.getUserByUsername(userName)  }
     }
 
     private fun userName(context: ActionContext): String = context.params(":username")
--- a/src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/UserController.kt
+++ b/src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/UserController.kt
@@ -10,7 +10,6 @@ class UserController(context: ControllerContext) : AbstractController(context)
 {
     override val urlComponent = "/users"
     override val endpoints = listOf(
-          //  BasicEndpoint("/", this::getUsers),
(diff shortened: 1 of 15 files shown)


src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/UserController.kt
```diff
@@ -0,0 +1,26 @@
...
+{
+    override val urlComponent = "/users"
+    override val endpoints = listOf(
+            SecuredEndpoint("/:username/", this::getUser, emptySet())
+    )
+
+    fun getUser(context: ActionContext): User
+    {
+        var userName = userName(context)
+        var user = repos.user().use { it.getUserByUsername(userName(context))  }
```
MartinEden: I assume you meant to use the userName variable here, rather than invoking the userName method again

src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/UserController.kt
```diff
@@ -0,0 +1,26 @@
...
+        if (user == null)
+            throw UnknownObjectError(userName, "Username")
```
MartinEden: I've been tending the follow the approach of making the repository endpoints return a non-nullable object, and then throwing the UnknownObjectException in the repository implementation. This means that if we use the same repository endpoint in multiple places we don't have to keep handling the null case.

`getUserByEmail` is a rather special case, because we want to throw a CredentialsException for that one.

src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/UserRepository.kt
```diff
@@ -1,8 +1,7 @@
 package org.vaccineimpact.api.app.repositories
 
-import org.vaccineimpact.api.models.permissions.User
-
 interface UserRepository : Repository
 {
-    fun getUserByEmail(email: String): User?
+    fun getUserByEmail(email: String): org.vaccineimpact.api.security.MontaguUser?
+    fun getUserByUsername(username: String): org.vaccineimpact.api.models.User?
 }
```
MartinEden: Now you've resolved the naming issue, you could go back to non-fully qualified class names here.

src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqUserRepository.kt
```diff
@@ -5,15 +5,15 @@ import org.vaccineimpact.api.app.repositories.UserRepository
...
-import org.vaccineimpact.api.models.permissions.UserProperties
-import org.vaccineimpact.api.models.permissions.ReifiedPermission
+import org.vaccineimpact.api.models.permissions.*
+
+import org.vaccineimpact.api.security.UserProperties
 
 class JooqUserRepository : JooqRepository(), UserRepository
 {
-    override fun getUserByEmail(email: String): User? {
+    override fun getUserByEmail(email: String): org.vaccineimpact.api.security.MontaguUser?
```
MartinEden: Also in this file it would be good to switch back to importing `MontaguUser`, rather than using the fully qualified name.

src/app/src/test/kotlin/org/vaccineimpact/api/tests/controllers/UserControllerTests.kt
```diff
@@ -0,0 +1,57 @@
...
+        val context = mock<ActionContext> {
+            on { hasPermission(any()) } doReturn true
+            on { params(":username") } doReturn userName
+        }
+
+        val controller = UserController(controllerContext)
+        assertThat(controller.getUser(context)).isEqualToComparingFieldByField(user)
+    }
+
+    @Test(expected = UnknownObjectError::class)
```
MartinEden: I've been tending to use `assertThatThrownBy { ... }.inInstanceOf(UnknownObjectError::class.java)` rather than `expected` attribute. My reason for preferring this is that then you are asserting that just that line (or block of code) under test throws the exception. Otherwise if it were thrown by some part of the test setup (unlikely in this case I know) that would also count as a passing test.

src/blackboxTests/src/test/kotlin/org/vaccineimpact/api/blackboxTests/UserTests.kt
```diff
@@ -0,0 +1,56 @@
...
+import org.vaccineimpact.api.models.permissions.PermissionSet
+import org.vaccineimpact.api.test_helpers.DatabaseTest
+
+class UserTests : DatabaseTest()
+{
+    @Test
+    fun `can get user by username`()
+    {
+        validate("/users/testuser") against "User" given {
+            it.addUserWithRoles("testuser", ReifiedRole("member", Scope.Specific("modelling-group", "group")))
```
MartinEden: It would be nice to also check the output for the role with a null scope prefix - e.g. the `user` role.

src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/UserController.kt
MartinEden: Although perhaps, since it's not clear from the repositories point of view that the endpoints are used for different things, we should bring `getUserByEmail` in line with everything else, and in `DatabasePasswordAuthenticator` catch the `UnknownObjectException` and rethrow it as a `CredentialsException`.

On other hand, I generally don't like using exceptions for logic like that. I regard the `UnknownObjectException` in a different light, because that one bubbles all the way to the top and is then handled by the general "turn this exception into an API error" mechanism.

So I leave it up to you whether to bring `getUserByEmail` in line, but either way I would like `getUserByUsername` to be like the other, ordinary repository endpoints.

littledice: That makes a lot of sense. I'll leave `getUserByEmail` as is, agree with not using exceptions to handle flow, but I'll rename to `getMontaguUserByEmail` just to emphasize the different return type.

src/blackboxTests/src/test/kotlin/org/vaccineimpact/api/blackboxTests/UserTests.kt
littledice: Is there a reason why `Scope.databaseScopeId` is nullable while `Scope.databaseScopeId` is not? Just thinking the inconsistency in the JSON for a global role ("scope_id" comes back as an empty string, "scope_prefix" as null) might be a little confusing to consumers of the API?

MartinEden: So the issue here is that we want them both to be nullable, but scope_id forms part of the primary key of `user_role`, so it can't be nullable. So we instead use the empty string for that. That's the technical reason in the background.

But what we should return in the API is another matter. I agree that it's kind of nasty. Maybe you could translate the empty string to a null?

MartinEden:
In case it's not clear, the only outstanding issue is the one about null vs empty string for scope ID - see the last comment above for more.

littledice:
Ok, modified the mapping method in `JooqUserRepository` to map empty `scopeId`s to null when populating roles. If you give this the thumbs up I'll merge.

PR #7, opened 2017-06-22
I346

littledice:
nb: this actually corresponds to issue 345, not issue 346!
https://vimc.myjetbrains.com/youtrack/issue/VIMC-345

### DIFF
--- a/src/app/src/main/kotlin/org/vaccineimpact/api/app/DirectActionContext.kt
+++ b/src/app/src/main/kotlin/org/vaccineimpact/api/app/DirectActionContext.kt
@@ -5,6 +5,7 @@ import org.pac4j.core.profile.ProfileManager
 import org.pac4j.sparkjava.SparkWebContext
 import org.vaccineimpact.api.app.errors.MissingRequiredPermissionError
 import org.vaccineimpact.api.app.security.montaguPermissions
+import org.vaccineimpact.api.db.tables.Permission
 import org.vaccineimpact.api.models.permissions.ReifiedPermission
 import spark.Request
 import spark.Response
@@ -30,6 +31,7 @@ open class DirectActionContext(private val context: SparkWebContext): ActionCont
 
     override fun hasPermission(requirement: ReifiedPermission)
             = permissions.any { requirement.satisfiedBy(it) }
+
     override fun requirePermission(requirement: ReifiedPermission)
     {
         if (!hasPermission(requirement))
--- a/src/app/src/main/kotlin/org/vaccineimpact/api/app/DirectActionContext.kt
+++ b/src/app/src/main/kotlin/org/vaccineimpact/api/app/DirectActionContext.kt
@@ -5,7 +5,6 @@ import org.pac4j.core.profile.ProfileManager
 import org.pac4j.sparkjava.SparkWebContext
 import org.vaccineimpact.api.app.errors.MissingRequiredPermissionError
 import org.vaccineimpact.api.app.security.montaguPermissions
-import org.vaccineimpact.api.db.tables.Permission
 import org.vaccineimpact.api.models.permissions.ReifiedPermission
 import spark.Request
 import spark.Response
--- a/src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/UserController.kt
+++ b/src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/UserController.kt
@@ -5,20 +5,43 @@ import org.vaccineimpact.api.app.controllers.endpoints.SecuredEndpoint
 import org.vaccineimpact.api.models.Scope
 import org.vaccineimpact.api.models.UserInterface
 import org.vaccineimpact.api.models.permissions.ReifiedPermission
+import org.vaccineimpact.api.models.permissions.RoleAssignment
 
 class UserController(context: ControllerContext) : AbstractController(context)
 {
     override val urlComponent = "/users"
-    override val endpoints = listOf(
-            SecuredEndpoint("/:username/", this::getUser, setOf("*/users.read"))
+    override val endpoints = listOf(SecuredEndpoint("/:username/", this::getUser, setOf("*/users.read"))
     )
 
     fun getUser(context: ActionContext): UserInterface
     {
-        var userName = userName(context)
+        val userName = userName(context)
 
-        return repos.user().use { it.getUserByUsername(userName) }
+        val roleReadingPermissions = context.permissions.filter { p -> p.name == "roles.read" }
+
+        if (!roleReadingPermissions.any())
+            return repos.user().use { it.getUserByUsername(userName) }
+
+        val userWithRoles = repos.user().use { it.getUserWithRolesByUsername(userName) }
+
+        userWithRoles.roles = userWithRoles.roles.filter { r ->
+            roleReadingPermissions.any {
+                p ->  p.scope.encompasses(parseScope(r))
+            }
+        }
+
+        return userWithRoles
     }
+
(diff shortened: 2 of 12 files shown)


src/app/src/main/kotlin/org/vaccineimpact/api/app/DirectActionContext.kt
```diff
@@ -5,6 +5,7 @@ import org.pac4j.core.profile.ProfileManager
 import org.pac4j.sparkjava.SparkWebContext
```
MartinEden: These changes could be reverted

src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqUserRepository.kt
```diff
@@ -67,6 +89,14 @@ class JooqUserRepository : JooqRepository(), UserRepository
     private fun caseInsensitiveUsernameMatch(username: String)
             = APP_USER.USERNAME.lower().eq(username.toLowerCase())
 
+    private fun permittedRoleReadingScopeIds(permissions: PermissionSet)
+            = permissions.filter { p -> p.name == "roles.read" }
+            .map { p -> p.scope.databaseScopeId }
+
+    private fun matchingOrGlobal(scopeId: String, record: Record) =
+            scopeId == record[USER_ROLE.SCOPE_ID]
+                    || scopeId.isNullOrEmpty()
```
MartinEden: I think we should be able to use `Scope.encompasses` here

src/app/src/test/kotlin/org/vaccineimpact/api/tests/controllers/UserControllerTests.kt
```diff
@@ -17,17 +20,20 @@ class UserControllerTests : ControllerTests<UserController>()
             = UserController(controllerContext)
 
     @Test
-    fun `getUserByUsername returns user`()
+    fun `getUser returns user`()
```
MartinEden: Why have you renamed this? Seemed fine before :)

src/blackboxTests/src/test/kotlin/org/vaccineimpact/api/blackboxTests/UserTests.kt
```diff
@@ -23,6 +23,48 @@ class UserTests : DatabaseTest()
...
+                        "name" to "Test User",
+                        "email" to "testuser@example.com",
+                        "last_logged_in" to null
+                )
+            })
+        }
+    }
+
+    @Test
+    fun `cannot get user without user reading permission`()
```
MartinEden: We don't need this test, as the fluent validation helper in the test above will use the `requiringPermissions` stanza and will automatically check that all those permissions are required. To prove this to yourself, remove the required permission from the controller and try running the test above.

src/databaseTests/src/test/kotlin/org/vaccineimpact/api/databaseTests/UserTests.kt
```diff
@@ -51,16 +54,16 @@ class UserTests : RepositoryTests<UserRepository>()
     fun `can retrieve user with any case username`()
     {
         given(this::addTestUser).check { repo ->
-            assertThat(repo.getUserByUsername("test.user")).isNotNull()
-            assertThat(repo.getUserByUsername("Test.User")).isNotNull()
+            assertThat(repo.getUserByUsername("test.user", PermissionSet()) as User).isNotNull()
+            assertThat(repo.getUserByUsername("Test.User", PermissionSet()) as User).isNotNull()
```
MartinEden: I should have picked this up sooner: now that the interface returns a non-nullable object, this check will always pass. We should instead be checking that it doesn't throw an exception. Also, you shouldn't have to cast to User to check that something is not null?

src/app/src/test/kotlin/org/vaccineimpact/api/tests/controllers/UserControllerTests.kt
littledice: the method it's testing is called 'getUser'

src/databaseTests/src/test/kotlin/org/vaccineimpact/api/databaseTests/UserTests.kt
littledice: Ah good point. Casting just to check the right object is returned. Can just do without the assertions then, just calling the method and casting to User without an error being thrown will be a test.

src/blackboxTests/src/test/kotlin/org/vaccineimpact/api/blackboxTests/UserTests.kt
littledice: ah cool

src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqUserRepository.kt
littledice: oh nice, I missed that

src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqUserRepository.kt
```diff
@@ -86,12 +108,19 @@ class JooqUserRepository : JooqRepository(), UserRepository
...
+            scopeId
+        }
 
         return RoleAssignment(
                 record[ROLE.NAME],
-                scopeId,
-                record[ROLE.SCOPE_PREFIX])
+                record[ROLE.SCOPE_PREFIX],
+                scopeId)
     }
```
littledice: note, just changed this constructor around so that prefix comes before id, as is the case everywhere else

src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/UserController.kt
```diff
@@ -2,22 +2,46 @@ package org.vaccineimpact.api.app.controllers
...
+        userWithRoles.roles = userWithRoles.roles.filter { r ->
+            roleReadingPermissions.any {
+                p ->  p.scope.encompasses(parseScope(r))
+            }
+        }
+
+        return userWithRoles
     }
 
+    private fun parseScope(role: RoleAssignment) =
```
MartinEden: It might be good to put this in `Scope`'s companion object. e.g. `fun fromRoleAssignment` so it can be reused. Also, rather than invoking `Scope.parse`, you may as well construct a `Scope.Specific` object directly.

src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqUserRepository.kt
```diff
@@ -61,6 +71,16 @@ class JooqUserRepository : JooqRepository(), UserRepository
         )
     }
 
+    private fun getUser(username: String): AppUserRecord
```
MartinEden: Consider rewriting this as:

```
private fun getUser(username: String) = dsl.fetchAny(APP_USER, caseInsensitiveUsernameMatch(username))
    ?: throw UnknownObjectError(username, "Username")
```

littledice: On the whole (for the sake of code readability) I prefer not to use ternary operators except for variable assignment

littledice:
Rre last comment, you mean Blackbox test? I have done exactly that if you take a look :)

PR #10, opened 2017-06-26
I344

littledice:
get all users without roles


MartinEden:
Just waiting on TeamCity before merging.

PR #12, opened 2017-06-26
i349: Create 'don't serialize if null' attribute for models

MartinEden:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-349

PR #13, opened 2017-06-27
i346: return users with roles if api user has required permission

littledice:
nb: should wait til after merging https://github.com/vimc/montagu-api/pull/12

also linked PR: https://github.com/vimc/montagu-webmodels/pull/4


src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/UserController.kt
```diff
@@ -14,31 +16,50 @@ class UserController(context: ControllerContext) : AbstractController(context)
...
 
+    private fun userName(context: ActionContext): String = context.params(":username")
 
-    fun getUsers(context: ActionContext): List<User>
-    {
-        return return repos.user().use { it.all().toList() }
-    }
+    private fun roleReadingScopes(context: ActionContext) = context.permissions
+            .filter { it.name == "roles.read" }
+            .map { it.scope }
```
littledice: I wonder if these private functions contain reusable logic/should be moved out into the roles class

MartinEden: Let's wait and see - we can always pull them out when we need them.

PR #17, opened 2017-07-05
Also load properties from /etc

MartinEden:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-368


MartinEden:
I wasn't able to write meaningful tests for this, as it required me to set up a file in `/etc/` for the app to read from - but the test doesn't have permission to write to that file. In the end, I think this is simple enough logic that we can get away without an automated test.

PR #18, opened 2017-07-05
Libsodium

littledice:
Use this wrapper around libsodium to hash/verify password: https://github.com/abstractj/kalium

Have added ticket to remove unused db field:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-374


scripts/install-libsodium.sh
```diff
@@ -0,0 +1,7 @@
+curl https://download.libsodium.org/libsodium/releases/libsodium-1.0.12.tar.gz \
+    | tar xz --directory lib/
+pushd lib/libsodium-1.0.12/
+./configure
```
littledice: so this will download the libsodium tar into the (non-existent) folder `/scripts/lib`... maybe `/lib` was intended?

littledice:
The build environment container now has libsodium installed for tests, but haven't got it configured inside the app container yet.

Dockerfile
```diff
@@ -15,6 +15,17 @@ RUN add-apt-repository \
...
+RUN apt-get install -y build-essential
+
+RUN curl https://download.libsodium.org/libsodium/releases/libsodium-1.0.12.tar.gz \
+    | tar xz 
+
+RUN cd libsodium-1.0.12/ \
+	&& ./configure \
+	&& make && make check \
+	&& make install
+
```
MartinEden: I know I said the opposite yesterday, but actually it does seem like it would be good to run `./scripts/install-libsodium.sh` here. I had forgotten that we would still need that script for developers installing on their own machine. At which point, if we can use the same script in both places that's good, as it means the script developers will use is being tested on TeamCity.

So the docker file becomes something like:
```
COPY ./scripts/install-libsodium.sh .
RUN ./install-libsodium.sh
```

src/security/src/main/kotlin/org/vaccineimpact/api/security/SodiumPasswordEncoder.kt
```diff
@@ -0,0 +1,27 @@
...
+import org.pac4j.core.credentials.password.PasswordEncoder
+
+class SodiumPasswordEncoder : PasswordEncoder
+{
+    private val passwordHelper = Password()
+
+    override fun encode(password: String)
+            = passwordHelper.hash(password.toByteArray(),
+            Encoder.HEX,
+            NaCl.Sodium.CRYPTO_PWHASH_SCRYPTSALSA208SHA256_OPSLIMIT_INTERACTIVE,
```
MartinEden: Maybe pull this out in the same way as `memoryLimit()`?

src/security/src/test/kotlin/SodiumTests.kt
```diff
@@ -0,0 +1,18 @@
...
+        val testPassword = "this is a password"
+
+        val hashedPw = encoder.encode(testPassword)
+        val decodedPw = encoder.matches(testPassword, hashedPw)
+
+        assert(decodedPw)
+
+    }
+
+}
```
MartinEden: This doesn't match the project code style - braces, etc. 

Also that final condition is confusing - `encoder.matches` returns a boolean, so the variable shouldn't be called `decodedPw`.

Maybe rewrite as: `assertThat(encoder.matches(testPassword, hashedPw)).isTrue()`

scripts/install-libsodium.sh
```diff
@@ -0,0 +1,15 @@
...
+    | tar xz --directory lib/
+
+pushd lib/libsodium-1.0.12/
+
+./configure
+make
+make check
+sudo make install
+
+popd
```
MartinEden: I updated this to add the `mkdir -p lib` command as otherwise it failed on my machine. While I was at it, I spaced it a bit more and added the `set -e` at the top.

src/security/src/main/kotlin/org/vaccineimpact/api/security/SodiumPasswordEncoder.kt
```diff
@@ -0,0 +1,27 @@
+package org.vaccineimpact.api.security
+
+import org.abstractj.kalium.NaCl
+import org.abstractj.kalium.crypto.Password
+import org.abstractj.kalium.encoders.Encoder
+import org.pac4j.core.credentials.password.PasswordEncoder
+
+class SodiumPasswordEncoder : PasswordEncoder
+{
+    private val passwordHelper = Password()
```
MartinEden: It would be nice to have a comment here explaining that `Password` uses scrypt with salsa, and that's why it makes sense to be passing in the `SCRYPTSALSA` constants for mem and ops limit. I'm not normally so careful to make things clear, but with this security stuff it's nice to be really clear.

littledice: added comment near the relevant methods

littledice:
I'm confused by that also, it didn't need `sudo` to run a `make install` command from the Docker file itself, but now the Docker file is calling an external shell script containing the same command, when I removed `sudo` it didn't seem to install anything.

littledice:
ok, I think this is all working now

cliApp.Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM montagu-api-build-environment
+
```
MartinEden: If I understand what you are doing here, this is the Dockerfile for the image that gets used to actually run the CLI in production. Whereas `cli.dockerfile` is the build environment image. In which case, this one should not extend from montagu-api-build-environment - we don't want our whole build env in production. You probably just want to extend from the same OpenJDK JRE image as the API `distDocker` task.

src/userCLI/build.gradle
```diff
@@ -36,11 +36,12 @@ if (hasProperty('docker_name')) {
     cli_docker_name = docker_name
 }
 
-distDocker {
+task cliDocker(type: Docker) {
```
MartinEden: I think this name is a bit unclear. How about `distCLIAsDockerImage`?

littledice:
Have simplified this; now creating a docker image with libsodium installed which both the build env and cli image inherit from. Using the distDocker task from the docker plugin as before.

PR #21, opened 2017-07-13
Add libsodium to API container

MartinEden:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-289


MartinEden:
As you can see, `master` fails at the moment: [TeamCity Link](http://teamcity.montagu.dide.ic.ac.uk:8111/viewLog.html?buildId=2016&tab=buildResultsDiv&buildTypeId=montagu_api_blackbox_tests).

But the direct descendant of that commit, which is the head of the `libsodium` branch passes: [TeamCity Link](http://teamcity.montagu.dide.ic.ac.uk:8111/viewType.html?buildTypeId=montagu_api_blackbox_tests&branch_montagu_MontaguApi=libsodium&tab=buildTypeStatusDiv)

MartinEden:
Nope, this isn't working. Spoke too soon.

MartinEden:
Okay, this is fixed now. Here's what happened:

* After we introduced the dependency on libsodium the black box tests started failing, but because the error was a build failure, not a test failure, and because the `run-blackbox-tests.sh` script didn't have `set -e` the error was not propagated to TeamCity. If we had been eagled-eyed we would have noticed that TeamCity was now reporting "Build passed" instead of "Tests passed 37".
* Then, we had the database schema change. I fixed that, but I didn't catch a few blackbox tests, because TeamCity was reporting everything as fine, and I trusted it.

Here's what I have done in this PR:

* Added libsodium both to the API container and to the blackbox tests environment.
* Fixed the test failures left over from the db schema change. This included relaxing the schema so that it doesn't require the fields we haven't implemented yet. I've added a note to https://vimc.myjetbrains.com/youtrack/issue/VIMC-396 to remind us to tighten the schema again once that's implemented.
* Added `set -e` to the blackbox tests script so errors get caught
* Added an additional check to the blackbox tests TeamCity configuration so that if the number of tests drops sharply it counts it as a build failure

PR #22, opened 2017-07-13
API should use configurable keypair for token signing

MartinEden:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-395

By default the API still generates a new keypair if none is provided. This makes testing (e.g. blackbox tests) much easier, and seems nice to me - otherwise the API container can't run the API by itself, which just seems awkward. It's not a security hole - it just means other containers can't automatically use the same public key.

There's also [this PR](https://github.com/vimc/montagu/pull/1) for the deployment tool that actually generates the keypair and injects it, but until we've talked through the deployment tool I'm not expecting you to review that. So I guess for now I'll just merge that one in once you've reviewed this PR.

### DIFF
--- a/scripts/run-blackbox-tests.sh
+++ b/scripts/run-blackbox-tests.sh
@@ -3,8 +3,7 @@ cert_tool_version=59657b2
 
 docker-compose pull
 docker-compose --project-name montagu up -d
-docker run docker.montagu.dide.ic.ac.uk:5000/montagu-cert-tool:$cert_tool_version gen-self-signed password > keystore
-./scripts/add-certificate-to-docker.sh keystore montagu_api_1
+docker exec montagu_api_1 touch /etc/montagu/api/go_signal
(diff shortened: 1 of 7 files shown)


scripts/run-blackbox-tests.sh
```diff
@@ -6,8 +6,8 @@ cert_tool_version=59657b2
 # Run API and DB
 docker-compose pull
 docker-compose --project-name montagu up -d
-docker run docker.montagu.dide.ic.ac.uk:5000/montagu-cert-tool:$cert_tool_version gen-self-signed password > keystore
-./scripts/add-certificate-to-docker.sh keystore montagu_api_1
```
MartinEden: This code was left over from when the API used SSL (instead of it being hidden behind a reverse proxy that uses SSL). It was generating a certificate that then wasn't being used.

PR #26, opened 2017-07-24
Use explicit DB version in API

MartinEden:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-432


generate-test-data.Dockerfile
```diff
@@ -3,4 +3,4 @@ FROM montagu-api-build-environment
 CMD ./gradlew :generateTestData \
   && ./user.sh add "Test User" test.user test.user@imperial.ac.uk password \
   && ./user.sh addRole test.user user \
-  && ./user.sh addRole test.user member modelling-group IC-Garske
+  && ./user.sh addUserToGroup test.user IC-Garske
```
MartinEden: Just an unrelated opportunistic refactor.

scripts/generate-test-data.sh
```diff
@@ -1,10 +1,16 @@
...
 
 # Setup the database on a named network
+db_version=$(<src/config/db_version)
 docker network create test-data
-docker run -d --rm --name db --network=test-data -p "8000:5432" docker.montagu.dide.ic.ac.uk:5000/montagu-db:master
+docker run -d --rm \
+  --name db \
+  --network=test-data \
+  -p "8000:5432" \
+  docker.montagu.dide.ic.ac.uk:5000/montagu-db:$db_version
```
MartinEden: The only difference here is changing from hardcoded `master` to `$db_version`; I just also moved everything on to multiple lines.

src/config/default.properties
```diff
@@ -5,7 +5,7 @@ db_username=vimc
 db_password=changeme
 db_driver=org.postgresql.Driver
 
-testdb_image=docker.montagu.dide.ic.ac.uk:5000/montagu-db:master
+testdb_image=docker.montagu.dide.ic.ac.uk:5000/montagu-db
```
littledice: tiny thing - why not put the colon on the end of here, rather than appending in the gradle task, makes code in build.gradle slightly simpler and might also make it clearer that this is an incomplete image path?

littledice:
approved with 1 comment - feel free to amend or not before merging

src/config/default.properties
MartinEden: This way the testdb_image is a valid identifier in its own right. I can see the advantages of both approaches; I think I'll leave it as is.

PR #27, opened 2017-07-26
I367

littledice:
Read in properties from external config file and give external properties priority.


MartinEden:
This isn't needed.

PR #28, opened 2017-08-02
Convert schema validation to Kotlin

MartinEden:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-455

This is just a straight re-implementation of the Python script, but with a better foundation of properly parsing the markdown as a structured document, rather than just reading it as a text file. This gives us more metadata that we can later (in a future ticket) use to improve the blackbox tests.


spec/spec.md
```diff
@@ -24,8 +24,7 @@ the standard response format, so that the `data` property holds the payload.
 ## Success
 Schema: [`Response.schema.json`](Response.schema.json)
 
-Example
-
+### Example
```
MartinEden: There are some changes here: The new validator insists that the examples are introduced with a header.

spec/spec.md
```diff
@@ -162,6 +160,7 @@ Schema: [`Users.schema.json`](Users.schema.json)
 Returns a particular user.
 
 Required permissions: `users.read`. Additionally, the `roles` section is not included unless the logged in user has the `roles.read` permission. If the user has `roles.read` in a specific scope, only roles with a matching scope are returned.
+
```
MartinEden: The new validator insists on a blank line between these declarations.

MartinEden:
To make sure I hadn't made any mistakes, I ran the two validators and diffed their output and made sure they were identical (other than whitespace). I then improved the output format of the new one to be more informative.

PR #29, opened 2017-08-02
I352

littledice:
Demographics spec for test download


spec/DemographicDataset.schema.json
```diff
@@ -0,0 +1,11 @@
+{
+	"id": "DemographicDataSet",
+	"type": "object",
+    "properties": {
```
MartinEden: Mixed spaces/tabs here?

spec/spec.md
```diff
@@ -1688,3 +1688,76 @@ Schema: [`TouchstoneOverview.schema.json`](TouchstoneOverview.schema.json)
...
+
+#### format
+Optional. A format to return the CSV in, either `wide` or `long`. Defaults to `long`.
+
+### Example
+`/touchstones/2017-op-1/demographics/tot_pop/?format=wide&gender=female`
+
+Total population:
+
+                                       "people"
```
MartinEden: What is this "people" header? Doesn't that break the CSV format?

weshinsley: It looks like just a helpful header to say what the numbers mean (besides deriving it from the filename)

CSV is very relaxed - readers just have to skip the right number of header rows. You could prefix with "","","", to encourage it to line up with the first data column.

Source-wise, it's the "name" column of the demographic_value_unit table, where id matches the demographic_value_unit field in the demographic_statistic_type table.

MartinEden:
Actually, the schema check if failing because we haven't merge in #28. Perhaps you could prioritise that, and it would make it clearer if this branch is okay.

spec/spec.md
littledice: Yup, exactly. Lining it up sounds like the right thing to do, will amend.

littledice:
Ok, there were some issues with the schema, fixed now. The demographic type codes in the db have underscores in them, but the `StandardIdentifier` schema only allows hyphens. So we should either
* change codes in db to match schema
* convert underscores to hyphens in the code
* use different schema for demographic stat type codes

We should raise a ticket to do one of these when this PR is merged.

PR #30, opened 2017-08-04
Create users via the API

MartinEden:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-363

Not ready for review yet.


MartinEden:
Okay, this is a big one. I think we should probably pair on reviewing it.

src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqUserRepository.kt
```diff
@@ -76,6 +77,15 @@ class JooqUserRepository(db: JooqContext) : JooqRepository(db), UserRepository
...
     }
 
+    override fun addUser(user: CreateUser)
+    {
+        dsl.newRecord(APP_USER).apply {
+            username = user.username
+            name = user.name
+            email = user.email
+        }.insert()
+    }
```
littledice: double check that a user with no pw cannot log in (add unit test)

src/app/src/main/resources/config.properties
```diff
@@ -10,3 +10,4 @@ token.issuer=${token_issuer}
 
 app.version=${app_version}
 app.port=${app_port}
+app.url=${app_url}
```
littledice: can you get this from the request?

src/blackboxTests/src/test/kotlin/org/vaccineimpact/api/blackboxTests/CreateUserTests.kt
```diff
@@ -0,0 +1,127 @@
...
+            "email" to "gandalf@example.com"
+    )
+
+    @Test
+    fun `can create user`()
+    {
+        val objectUrl = "/users/gandalf.grey/"
+
+        validate("/users/", HttpMethod.post) against "CreateUser" sending {
+            postData.toJsonObject()
```
littledice: can we avoid overloading 'against' here

src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqUserRepository.kt
MartinEden: Fixed

src/app/src/main/resources/config.properties
MartinEden: It turns out yes, but as I suspected it gets messed up by the reverse proxy.

src/blackboxTests/src/test/kotlin/org/vaccineimpact/api/blackboxTests/CreateUserTests.kt
MartinEden: Replaced it with `withRequestSchema`.

MartinEden:
Ready for review again.

PR #34, opened 2017-08-10
Make database tests faster

MartinEden:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-474

This takes the database tests from about 60 seconds to about 30 seconds on TeamCity. Most of the TeamCity task time is spent downloading dependencies and compiling code, so it doesn't help there that much. But it helps for when you want to quickly run the tests locally.


src/build.gradle
```diff
@@ -66,18 +66,27 @@ task startDatabase(dependsOn: ':loadPropertiesTask') {
...
         exec {
-            commandLine 'docker', 'run', '-d', '--rm', '-p', portBinding, '--name', 'montagu-api-test_database', image
+            commandLine 'docker', 'run',
+                    '-d',
+                    '--rm',
+                    '-p', portBinding,
+                    '--name', container,
+                    "--entrypoint", "start-with-config.sh",
+                    image,
+                    "/postgresql.test.conf"
```
MartinEden: You can see the script it is running here: https://github.com/vimc/montagu-db/blob/master/start-with-config.sh

It just runs the normal entrypoint for the container, but first it concatenates the given conf file (in this case `/postgresql.test.conf`) onto the end of the main postgresql.conf file.

PR #37, opened 2017-08-10
update demography spec to allow for mutliple sources

littledice:


littledice:
Also https://github.com/vimc/montagu-webmodels/pull/7

PR #41, opened 2017-08-15
I458 - GET endpoint for CSV of demographic data for a touchstone

littledice:
* Endpoint to get metadata plus CSV of a demographic dataset by source, touchstone and type parameters, gender and variants hard-coded
* CSV only and onetime token CSV only endpoint 

At the moment the CSV file headers include 'age', 'value'. We want to use the age interpretation and unit for individual statistic types to populate these headers. Have opened a separate ticket for this: https://vimc.myjetbrains.com/youtrack/issue/VIMC-509


littledice:
Not ready for review yet

littledice:
Ready for review

littledice:
also https://github.com/vimc/montagu-webmodels/pull/8

spec/DemographicData.csvschema.json
```diff
@@ -0,0 +1,9 @@
+{
+    "columns": {
+        "country": "String",
+        "age": "String",
+        "year": "Int",
+        "gender": "String",
+        "value": "Decimal"
+    }
+}
```
MartinEden: Is a string age range really the best way of communicating the age range? Wouldn't two columns `ageFirst` and `ageLast`, as in the coverage data, be better? Otherwise anyone consuming this programmatically has to parse the "this - that" format. Also, it's a little unclear whether the upper age bound is inclusive or exclusive.

Sorry if this is something that has been discussed to death with Wes and Tini already.

src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/TouchstoneController.kt
```diff
@@ -55,6 +62,25 @@ class TouchstoneController(context: ControllerContext) : AbstractController(cont
         return repo.getDemographicStatisticTypes(touchstone.id)
     }
 
+    fun getDemographicDataAndMetadata(context: ActionContext, repo: TouchstoneRepository):
+            SplitData<DemographicDataForTouchstone, DemographicRow>
+    {
+        val touchstone = touchstone(context, repo)
+        val source = context.params(":source-code")
+        val type = context.params(":type-code")
+        logger.info("$type, $source, ${touchstone.id}")
```
MartinEden: Don't need this debug line in now, presumably?

src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/TouchstoneController.kt
```diff
@@ -55,6 +62,25 @@ class TouchstoneController(context: ControllerContext) : AbstractController(cont
...
+        return repo.getDemographicDataset(type, source, touchstone.id)
+    }
+
+    fun getDemographicData(context: ActionContext, repo: TouchstoneRepository): DataTable<DemographicRow>
+    {
+        val data = getDemographicDataAndMetadata(context, repo)
+        val metadata = data.structuredMetadata
+        val filename = "${metadata.touchstone.id}_${context.params(":source-code")}_${metadata.demographicData.id}.csv"
```
MartinEden: Might be more readable if you first pulled out the source code into a variable, to keep the interpolation simpler.

src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/TouchstoneController.kt
```diff
@@ -55,6 +62,25 @@ class TouchstoneController(context: ControllerContext) : AbstractController(cont
...
+        context.addResponseHeader("Content-Disposition", """attachment; filename="$filename"""")
```
MartinEden: Maybe refactor so we have a reusable method `addAttachmentHeader(filename: String)` and share it with the other CSV download?

src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqTouchstoneRepository.kt
```diff
@@ -208,4 +329,12 @@ class JooqTouchstoneRepository(
...
             record[COVERAGE.COVERAGE_]
     )
+
+    fun mapDemographicRow(record: Record) = DemographicRow(
+            record[DEMOGRAPHIC_STATISTIC.COUNTRY],
+            "${record[DEMOGRAPHIC_STATISTIC.AGE_FROM]} - ${record[DEMOGRAPHIC_STATISTIC.AGE_TO]}",
+            record[DEMOGRAPHIC_STATISTIC.YEAR],
+            record[GENDER.NAME],
+            record[DEMOGRAPHIC_STATISTIC.VALUE]
+    )
```
MartinEden: I haven't traced through every part of these SQL queries in detail. Question: Did you do any performance testing on these in the end? i.e. against the real data? Because I know you spent some time thinking about query optimisation. It seems nice and fast with the test data, but that's very small.

littledice: The main query to retrieve the rows is written in the most naive way possible, as I wanted to be really clear I had got the logic right first. I think it can and should be optimized though, so lets open a ticket for that.

spec/DemographicData.csvschema.json
littledice: This age range field modeled on the format of the unwpp data, and both bounds are inclusive (resulting in the strange looking values 0-0, 1-1, 2-2 when the age step is 1!) But that's not to say it's the best way, might be worth running it by Wes again.

src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqTouchstoneRepository.kt
littledice: https://vimc.myjetbrains.com/youtrack/issue/VIMC-514

spec/DemographicData.csvschema.json
littledice: ok, updated to use integer columns

PR #47, opened 2017-08-17
pass query parameters through to one time tokens

littledice:
also changed methods to expect long form gender names rather than single char codes, so "F" -> "female" etc


src/app/src/test/kotlin/org/vaccineimpact/api/tests/controllers/AbstractControllerTests.kt
```diff
@@ -19,9 +28,34 @@ class AbstractControllerTests : MontaguTests()
...
-        val c = Controller()
+        val c = Controller(ControllerContext("/v6", mock(), mock()))
         assertThat(c.buildPublicUrl("/fragment/")).endsWith(
             "/v6/test/fragment/"
         )
     }
+
+    @Test
+    fun `can get onetime link token`()
+    {
```
littledice: I moved this test out of the abstract base class `ControllerTests<T>` as it was being run multiple times (once for each implementation of the class)

src/build.gradle
```diff
@@ -122,7 +122,7 @@ subprojects {
     task copyTestConfig(type: Copy, dependsOn: ':loadPropertiesTask') {
         from 'src/test/resources'
         include '*'
-        into 'build/classes/test/'
+        into 'out/test/classes/'
         expand(rootProject.ext.props)
```
littledice: Even after I set the output directories back to the old paths, IntelliJ changed them back when I reloaded the project! But if we're both using the latest IntelliJ may as well hard-code these to the newly correct values in lieu of a proper fix.

src/app/src/main/kotlin/org/vaccineimpact/api/app/OneTimeLink.kt
```diff
@@ -47,11 +51,26 @@ data class OneTimeLink(val action: OneTimeAction, val payload: Map<String, Strin
...
-            return OneTimeLink(action, payload)
+
+            val queryParams =
+                    if (rawQueryParams.isNullOrEmpty())
+                    {
+                        mapOf()
+                    }
+                    else
+                    {
+                        rawQueryParams!!
```
MartinEden: If you replace the if statement with `if (rawQueryParams == null || rawQueryParams == "")` then you won't need the `!!` null coercion.

src/app/src/main/kotlin/org/vaccineimpact/api/app/OneTimeLink.kt
```diff
@@ -47,11 +51,26 @@ data class OneTimeLink(val action: OneTimeAction, val payload: Map<String, Strin
...
+                                .split("&")
+                                .map { it.split('=') }
+                                .associateBy({ it[0] }, { it[1] })
```
MartinEden: This duplicates the decoding performed for the payload, right? If so, please factor out into a function.

src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqTouchstoneRepository.kt
```diff
@@ -259,9 +259,9 @@ class JooqTouchstoneRepository(
                 .join(DEMOGRAPHIC_STATISTIC_TYPE)
                 .on(DEMOGRAPHIC_STATISTIC_TYPE.ID.eq(DEMOGRAPHIC_STATISTIC.DEMOGRAPHIC_STATISTIC_TYPE))
 
-        // if gender is not applicable for this statistic type, ignore passed genderCode parameter and match on "B"
+        // if gender is not applicable for this statistic type, ignore passed gender parameter and match on "B"
         val genderMatchesOrShouldBeDefault = (GENDER.CODE.eq("B").andNot(DEMOGRAPHIC_STATISTIC_TYPE.GENDER_IS_APPLICABLE))
-                .or(GENDER.CODE.eq(genderCode))
+                .or(GENDER.NAME.eq(gender))
```
MartinEden: This needs to stay as `GENDER.CODE`

PR #50, opened 2017-08-21
I529

littledice:
Recreating (https://github.com/vimc/montagu-api/pull/49) as the old PR wasn't updating to reflect the last merge to master


src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/PasswordController.kt
```diff
@@ -1,26 +1,69 @@
...
+    }
+
+    fun requestLink(context: ActionContext, repo: UserRepository): String
+    {
+        val address = context.queryParams("email")
+                ?: throw MissingRequiredParameterError("email")
+        val user = repo.getMontaguUserByEmail(address)
+        if (user != null)
+        {
+            val token = getSetPasswordToken(user.username, context)
```
littledice: The architecture is getting a bit confusing here - really the endpoint that uses this method is not a 'one repo' endpoint because it depends on both the user and the token repository, the latter being accessed directly from the controller context, the former passed in to the method.

The failing unit tests are because the token repo has not been mocked on the mock controller context, but I can see why that was missed because there isn't a clear dependency injection pattern! Overall I do think this demonstrates the limitation of the 'one repo endpoint' and that we should use a consistent depedency injection pattern.

 I'm tempted to say given time constraints   just fix the tests and open a ticket for refactoring, but the danger is then that the refactor will never be a top priority. Let's discuss in person.

src/emails/src/main/resources/templates/password-set.html
```diff
@@ -0,0 +1,23 @@
...
+        <p>
+            To change your password on Montagu please click
+            <a href="https://montagu.vaccineimpact.org/admin/set_password?token={{token}}">here</a>.
+        </p>
+        <p>
+            If you did not request a password change, or you no longer want to
+            change your password, please just ignore this email.
+        </p>
+        <p>Have a great day!</p>
+    </body>
```
littledice: Lol

littledice:
Some failing unit tests - see comment on code.

PR #52, opened 2017-08-24
Optimisequeries

littledice:


src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqTouchstoneRepository.kt
```diff
@@ -171,36 +170,56 @@ class JooqTouchstoneRepository(
...
+        val statsInTouchstoneSources = dsl.selectDistinct(DEMOGRAPHIC_STATISTIC.DEMOGRAPHIC_SOURCE,
+                DEMOGRAPHIC_STATISTIC.DEMOGRAPHIC_STATISTIC_TYPE.`as`("typeId"),
+                DEMOGRAPHIC_STATISTIC.COUNTRY.`as`("country"), field(name("s", "sourceCode"), String::class.java))
+                .from(DEMOGRAPHIC_STATISTIC)
+                .join(table(name("s")))
+                .on(DEMOGRAPHIC_STATISTIC.DEMOGRAPHIC_SOURCE.eq(field(name("s", "id"), Int::class.java)))
+
+        val countriesInTouchstone = countriesInTouchstone(touchstoneId)
+
+        return dsl.with("s").`as`(touchstoneSources)
```
MartinEden: Maybe we could call the subquery `"sources"` rather than `"s"`? I know this isn't new in this branch, but reviewing it the second time I'm feeling less overwhelmed and being a bit more opinionated ;)

Similarly, maybe `"sts"` could become `"statsInSources"`?

Finally, to make sure these don't get out of sync, would it be worth pulling these strings into consts?

e.g. 
```
val TOUCHSTONE_SOURCES = "touchstoneSources"
val STATS_IN_SOURCES = "statsInTouchstoneSources"
...
field(name(STATS_IN_SOURCES, "sourceCode")
...
etc.
```

src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqTouchstoneRepository.kt
```diff
@@ -69,8 +68,8 @@ class JooqTouchstoneRepository(
 
         val referenceRecord = records.firstOrNull()
 
-        val source = referenceRecord?.get(DEMOGRAPHIC_SOURCE.NAME)
-        val gender =  referenceRecord?.get(GENDER.NAME)
+        val source = referenceRecord?.get(field(name("s", "sourceCode"), String::class.java))
+        val gender = referenceRecord?.get(GENDER.NAME)
```
MartinEden: Would it be worth writing a generic inline function to make this code a little more readable?

e.g.

```
inline fun <reified T: Any?> Record.get(field: Field<*>): T = this.get(field, T::class.java)
```

Then you can replace the `val source = ...` line with:

```
referenceRecord?.get<String>(field(name("s", "sourceCode")))
```

Which at least removes the need for the `::class.java`

PR #56, opened 2017-09-13
I661 - Include current burden estimate in responsibility endpoint

littledice:
Relies on this change to the db:
https://github.com/vimc/montagu-db/pull/9/files


src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqModellingGroupRepository.kt
```diff
@@ -142,9 +141,59 @@ class JooqModellingGroupRepository(
...
+            }
+            else
+            {
+                ResponsibilityStatus.VALID
+            }
+        }
+        return Responsibility(scenario, status, problems, burdenEstimate)
+    }
+
+    private fun getCurrentBurdenEstimate(responsibilityId: Int): BurdenEstimateSet?
```
littledice: this should really be a public function on the burden estimate repo, just waiting on i121 so that that repo exists!

MartinEden: Cool, we can move that across in a bit. In the mean time, could you rename `burdenEstimate` to `burdenEstimateSet` throughout this section?

src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqModellingGroupRepository.kt
```diff
@@ -142,9 +141,59 @@ class JooqModellingGroupRepository(
...
+    {
+        val records = dsl.select(BURDEN_ESTIMATE_SET_PROBLEM.PROBLEM)
+                .select(BURDEN_ESTIMATE_SET.UPLOADED_ON, BURDEN_ESTIMATE_SET.ID)
+                .from(BURDEN_ESTIMATE_SET)
+                .leftJoin(BURDEN_ESTIMATE_SET_PROBLEM)
+                .on(BURDEN_ESTIMATE_SET_PROBLEM.BURDEN_ESTIMATE_SET.eq(BURDEN_ESTIMATE_SET.ID))
+                .join(RESPONSIBILITY)
+                .on(RESPONSIBILITY.CURRENT_BURDEN_ESTIMATE_SET.eq(BURDEN_ESTIMATE_SET.ID))
+                .where(RESPONSIBILITY.ID.eq(responsibilityId))
+                .fetch()
```
MartinEden: I *think* that what you're doing here is complicated enough you can't use the `joinPath` helpers I wrote, but just to check: Are you aware of the `joinPath` helpers and are you comfortable using them? If not, I can give you a quick briefing on them.

littledice: Have changed to use the helper for the first join, the second one I think needs specifying precisely as there are 2 possible join paths from `responsibility` to `burden_estimate_set`

PR #60, opened 2017-09-18
Handle transactions properly

MartinEden:
This required a pretty big rewrite, but I think the resulting code is actually a fair bit simpler.

Here's basically what's changed: Previously we had the `Repositories` object, that contained a factory for each repository, which would provide each one with their own database connection. Now we have a separate `RepositoryFactory` class that will make an instance of the `Repositories` object, which now contains actual instantiated repositories, all sharing a single database connection and transaction.

This has resulted in the code getting simpler in most places. The places where it has gotten more complicated are places where we were kind of doing things in a fudged way anyway before; the new approach forces us to always handle it correctly.


littledice:
Merging but would like to refactor this if I get the time.

PR #80, opened 2017-10-11
Allow developers to disable authentication on the API in development

MartinEden:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-795

### DIFF
--- a/src/config/default.properties
+++ b/src/config/default.properties
@@ -15,6 +15,7 @@ testdb_container=montagu-api-test-db
 
 docker_host=tcp://localhost:2375
 
+app_auth=true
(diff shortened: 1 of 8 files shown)


src/config/default.properties
```diff
@@ -15,6 +15,7 @@ testdb_container=montagu-api-test-db
 
 docker_host=tcp://localhost:2375
 
+app_auth=true
```
littledice: Is it worth also setting app_auth=true in the docker.properties file? The latter being production values, these being default or local ones. An extra guard against app_auth being set to true in a deploy!

PR #81, opened 2017-10-13
Refactor demographic stat queries to use new data structure

littledice:
* Simplified query to get available statistic types for a touchstone

* Re-wrote query to get actual stats removing all references to `touchstone_demographic_source`

* Quite a bit of renaming, to make models/code reflect the database structure more closely 

* Updated spec for  `GET /touchstones/{touchstone-id}/demographics/`. Effectively this should now return an array which reflects the `touchstone_demographic_dataset` table; each item in the array corresponds to one row in the `demographic_dataset` table (so should have a single `source` field rather than an array of sources)


spec/spec.md
```diff
@@ -1713,15 +1713,17 @@ Schema: [`Demographics.schema.json`](Demographics.schema.json)
     [{ 
         "id" : "as-fert",
         "name": "Age-specific fertility",
-        "sources" : ["unwpp2015", "unwpp2017"],
-        "countries" : ["AFG"],
+        "source" : "unwpp2015",
+        "sources": ["unwpp2015"],
+        "countries": [],
```
littledice: Idea is to remove `countries` and `sources` entirely. Leaving them in for now so that this is a non-breaking change. Once the portals have been updated to remove all references to these can remove from model: https://vimc.myjetbrains.com/youtrack/issue/VIMC-808

spec/DemographicDataForTouchstone.schema.json
```diff
@@ -1,10 +1,10 @@
 {
-    "id": "DemographicDatasetForTouchstone",
+    "id": "DemographicDataForTouchstone",
     "type": "object",
```
littledice: Renaming as `demographic_dataset` is now the name of a database table, whereas this maps onto the `demographic_statistic` table.

spec/spec.md
MartinEden: Could we add this comment to the spec itself, so our intention isn't only preserved in this PR?

spec/spec.md
```diff
@@ -1713,15 +1713,17 @@ Schema: [`Demographics.schema.json`](Demographics.schema.json)
...
         "gender_is_applicable": false
      },
      { 
         "id" : "tot-pop",
         "name" : "Total population",        
-        "sources" : ["unwpp2015"],
-        "countries" : ["AFG"],
+        "source" : "unwpp2015",
+        "sources": ["unwpp2015"],
+        "countries": [],
```
MartinEden: Also add comment here about these fields being deprecated.

src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqTouchstoneRepository.kt
```diff
@@ -64,10 +64,10 @@ class JooqTouchstoneRepository(dsl: DSLContext, private val scenarioRepository:
 
         val referenceRecord = records.firstOrNull()
 
-        val source = referenceRecord?.getField<String>(name(TOUCHSTONE_SOURCES, "sourceCode"))
+        val source = referenceRecord?.getField<String>(name(TOUCHSTONE_SOURCES, "code"))
```
MartinEden: Why has this changed?

littledice: no good reason!

PR #82, opened 2017-10-16
Optimize demographics query and CSV writing

MartinEden:
Two separate things in this PR, so could be better, but at least I remembered to create a PR at this point before going further!

1. The demographics data query.
2. Don't have special escaping rules for `NA`, instead use `<NA>` as our null marker. This allows us to remove `MontaguCSVWriter`, which was a big slowdown.


littledice:
Nice red diff and manageable sized PR :)

PR #85, opened 2017-10-18
Stream CSV data to browser

MartinEden:
This is part of the https://vimc.myjetbrains.com/youtrack/issue/VIMC-781 work.


src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/ModellingGroupController.kt
```diff
@@ -116,27 +116,6 @@ open class ModellingGroupController(context: ControllerContext)
...
-
-        request.raw().getPart("file").inputStream.bufferedReader().use {
-
-            // Then add the burden estimates
-            val data = DataTableDeserializer.deserialize(it.readText(), BurdenEstimate::class, Serializer.instance).toList()
-            return uploadBurdenEstimates(data, estimateRepository, context, path)
-        }
-
-    }
-
```
MartinEden: I didn't mean to include these changes here - sorry. However, all I have done is renamed the three burden estimate functions (as we discussed, they had confusing names) and moved them to all be next to each other.

src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/TouchstoneController.kt
```diff
@@ -78,8 +80,9 @@ class TouchstoneController(context: ControllerContext) : AbstractController(cont
         val metadata = data.structuredMetadata
         val source = context.params(":source-code")
         val gender = context.queryParams("gender")?: "both"
-        val filename = "${metadata.touchstone.id}_${source}_${metadata.demographicData.id}_${gender}.csv"
+        val filename = "${metadata.touchstone.id}_${source}_${metadata.demographicData.id}_$gender.csv"
```
MartinEden: Just removed unnecessary braces.

src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/endpoints/Endpoint.kt
```diff
@@ -39,8 +37,6 @@ data class Endpoint<TRoute>(
 
     override fun transform(x: Any) = when (x)
     {
-        is SplitData<*, *> -> x.serialize(Serializer.instance)
-        is DataTable<*> -> x.serialize(Serializer.instance)
         is AuthenticationResponse -> Serializer.instance.gson.toJson(x)!!
         else -> Serializer.instance.toResult(x)
```
MartinEden: It would have been great if we could have done the streaming here, completely auto-magically. Unfortunately, the response transformer does not have access to the output stream.

src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/endpoints/StreamedEndpoint.kt
```diff
@@ -0,0 +1,16 @@
...
+
+import org.vaccineimpact.api.app.ActionContext
+import org.vaccineimpact.api.app.serialization.Serializer
+import org.vaccineimpact.api.app.serialization.StreamSerializable
+
+fun <TRepository> ((ActionContext, TRepository) -> StreamSerializable).streamed()
+        : (ActionContext, TRepository) -> Unit
+{
+    return { context, repo -> stream(this(context, repo), context) }
+}
```
MartinEden: We can add more overloads later, if necessary.

src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/ModellingGroupController.kt
```diff
@@ -160,30 +139,40 @@ open class ModellingGroupController(context: ControllerContext)
...
+        request.raw().getPart("file").inputStream.bufferedReader().use {
+
+            // Then add the burden estimates
+            val data = DataTableDeserializer.deserialize(it.readText(), BurdenEstimate::class, Serializer.instance).toList()
+            return saveBurdenEstimates(data, estimateRepository, context, path)
+        }
     }
 
-    private fun getValidResponsibilityPath(context: ActionContext, estimateRepository: BurdenEstimateRepository): ResponsibilityPath
+    open fun addBurdenEstimates(context: ActionContext, estimateRepository: BurdenEstimateRepository): String
```
littledice: not introduced in this PR, but is there a reason for this to be an open function?

src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/endpoints/Endpoint.kt
littledice: So what happens when Spark tries to transform a streamed response? Does it just never hit this method because `StreamedEndpoint.strream()` doesn't return anything?

MartinEden: I think it does transform the response (which is just `kotlin.Unit`) but because we have already started writing to the response stream this output is discarded.

src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/ModellingGroupController.kt
MartinEden: I think for mocking in the tests.

MartinEden:
A couple of the onetime link tests were using `ModellingGroupController.getCoverageData` as an example one time action. Now that this involves streaming it's not the simplest action, so rather than trying to mock all the streaming stuff I just moved to using the "set password" action instead.

src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/ModellingGroupController.kt
```diff
@@ -193,13 +182,24 @@ open class ModellingGroupController(context: ControllerContext)
         val id = estimateRepository.addBurdenEstimateSet(
                 path.groupId, path.touchstoneId, path.scenarioId,
                 data,
-                uploader = context.username!!,
+                uploader = context.username ?: "testing",
```
littledice: Where is this "testing" username needed? Can it be provided via a mock instead of in the code?

src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/endpoints/Endpoint.kt
littledice: Ok, I guess no harm in letting it do that.

src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/ModellingGroupController.kt
MartinEden: Oops, that's just a mistake left in from testing. Good catch!

PR #86, opened 2017-10-18
flexible datatable 

littledice:
One models change: https://github.com/vimc/montagu-webmodels/pull/18

### DIFF
--- a/src/app/src/main/kotlin/org/vaccineimpact/api/app/serialization/FlexibleDataTable.kt
+++ b/src/app/src/main/kotlin/org/vaccineimpact/api/app/serialization/FlexibleDataTable.kt
@@ -1,15 +1,10 @@
 package org.vaccineimpact.api.app.serialization
 
-import com.opencsv.CSVWriter
 import org.vaccineimpact.api.ContentTypes
 import org.vaccineimpact.api.models.helpers.FlexibleProperty
-import java.io.OutputStream
 import kotlin.reflect.KClass
-import kotlin.reflect.KFunction
 import kotlin.reflect.KParameter
-import kotlin.reflect.full.declaredMemberProperties
 import kotlin.reflect.full.findAnnotation
-import kotlin.reflect.full.primaryConstructor
 
 class FlexibleDataTable<T : Any>(data: Iterable<T>,
                                  private val flexibleHeaders: Iterable<Any>,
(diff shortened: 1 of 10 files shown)


src/app/src/main/kotlin/org/vaccineimpact/api/app/serialization/FlexibleDataTable.kt
```diff
@@ -0,0 +1,108 @@
...
+    {
+        val basicValuesAsArray = headers
+                .map { it.property.get(line) }
+                .map { serializer.serializeValue(it) }
+                .toTypedArray()
+
+        val flexibleValuesAsArray = flexibleHeaders
+                .map { getFlexibleValue(it, line) }
+                .map { serializer.serializeValue(it) }
+                .toTypedArray()
```
MartinEden: Feels like there is some duplication here. Why not:

```
val values = headers.map { it.property.get(line) } 
           + flexibleHeaders.map { getFlexibleValue(it, line) }
return values
     .map { serializer.serializeValue(it) }
     .toTypedArray()
```

src/app/src/main/kotlin/org/vaccineimpact/api/app/serialization/FlexibleDataTable.kt
```diff
@@ -0,0 +1,108 @@
+package org.vaccineimpact.api.app.serialization
```
MartinEden: There is quite a lot of code here that is shared with `DataTable`. I think it would be nicer to extend that class and just override the necessary bits? e.g. Both classes construct and flush the writers identically.

src/app/src/main/kotlin/org/vaccineimpact/api/app/serialization/Serializer.kt
```diff
@@ -79,6 +80,13 @@ open class Serializer
         return text.toLowerCase().replace('_', '-')
     }
 
+    fun serializeValue(value: Any?) = when (value)
```
MartinEden: Maybe in this context it needs a more explicit name? e.g. `serializeValueForCSV`

src/app/src/test/kotlin/org/vaccineimpact/api/tests/serialization/FlexibleDataTableTests.kt
```diff
@@ -0,0 +1,64 @@
...
+        val data = listOf(
+                ABC("g", "h", mapOf("extra1" to "i", "extra2" to "j")),
+                ABC("x", "y", mapOf("extra1" to "z", "extra2" to "w"))
+        )
+        val table = FlexibleDataTable.new<ABC>(data, listOf("extra1", "extra2"))
+
+        Assertions.assertThat(serialize(table)).isEqualTo("""a,b,extra1,extra2
+g,h,i,j
+x,y,z,w""")
+    }
```
MartinEden: A test for the case where there is a mismatch between the extra headers supplied in the constructor, and in the individual rows?

MartinEden:
Nice!

PR #87, opened 2017-10-18
Allow download of wide format demographic data

littledice:
Not to be reviewed until https://github.com/vimc/montagu-api/pull/86 is merged, and then master merged into this branch.


spec/WideDemographicData.csvschema.json
```diff
@@ -0,0 +1,12 @@
+{
+    "columns": {
+        "country_code_numeric": "Int",
```
MartinEden: Why `country_code_numeric`? Is this a thing in the existing long format?

src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/TouchstoneController.kt
```diff
@@ -71,21 +72,74 @@ class TouchstoneController(context: ControllerContext) : AbstractController(cont
...
+        return if (format == "wide")
+        {
+            SplitData<DemographicDataForTouchstone, WideDemographicRow>(splitData.structuredMetadata,
+                    getWideDemographicDatatable(splitData.tableData.data))
+        }
+        else
+        {
+            SplitData<DemographicDataForTouchstone, LongDemographicRow>(splitData.structuredMetadata,
+                    splitData.tableData)
+        }
```
MartinEden: Be nice to throw an exception if the format is a unknown value?

MartinEden: Also, are the explicit type parameters necessary?

src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/TouchstoneController.kt
```diff
@@ -71,21 +72,74 @@ class TouchstoneController(context: ControllerContext) : AbstractController(cont
...
     }
 
-    fun getDemographicData(context: ActionContext, repo: TouchstoneRepository): DataTable<DemographicRow>
+    private fun getWideDemographicDatatable(data: Iterable<LongDemographicRow>):
+            FlexibleDataTable<WideDemographicRow>
+    {
+        val groupedRows = data
+                .groupBy {
+                    "${it.countryCode} ${it.ageFrom} " +
+                            "${it.ageTo}"
```
MartinEden: Might this be more readable as `.groupBy { Triple(it.countryCode, it.ageFrom, it.ageTo) }`, without the string interpolation gumph? Because `Triple` is a `data class`, it exhibits value semantics for `equals`, which I think it what you are shooting for here?

src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqTouchstoneRepository.kt
```diff
@@ -10,6 +10,7 @@ import org.vaccineimpact.api.app.repositories.ScenarioRepository
 import org.vaccineimpact.api.app.repositories.SimpleDataSet
```
MartinEden: There's a number of changes in this file that I think are just the result of you refactoring and then factoring back again. By all means keep changes that you think are improving readability, but if there are some that are more or less incidental, maybe you revert them so that we get a clearer record of what's actually changed here - and to make it easier to review! I **think** very little has changed here, but I'm not certain.

spec/WideDemographicData.csvschema.json
littledice: This was added to the spec here: https://vimc.myjetbrains.com/youtrack/issue/VIMC-753

src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqTouchstoneRepository.kt
littledice: Ah yes, you're right, I'll try revert

littledice:
Have merged in master now so diff should be a little clearer!

src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/TouchstoneController.kt
```diff
@@ -72,21 +75,71 @@ class TouchstoneController(context: ControllerContext) : AbstractController(cont
...
+        val format = context.queryParams("format")
+
+        val splitData = repo.getDemographicData(type, source, touchstone.id, gender ?: "both")
+
+        val tableData = when(format){
+
+            "wide" -> getWideDemographicDatatable(splitData.tableData.data)
+            "long" -> splitData.tableData
+            else -> throw BadRequest("Format '$format' not a valid csv format. Available formats are 'long' " +
+                    "and 'wide'.")
```
MartinEden: Sorry to keep being picky about this but: We should still probably support a default value if the user doesn't choose a format at all, right? Does `queryParams` return `null` if the parameter hasn't been specified. If so the elvis operator (`?:`) would be the quick fix.

spec/WideDemographicData.csvschema.json
MartinEden: :+1:

src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/TouchstoneController.kt
littledice: Oh yeah, there's also a Teamcity test failing bc of this, you're totally right

src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/TouchstoneController.kt
```diff
@@ -79,10 +79,11 @@ class TouchstoneController(context: ControllerContext) : AbstractController(cont
 
         val splitData = repo.getDemographicData(type, source, touchstone.id, gender ?: "both")
 
-        val tableData = when(format){
+        val tableData = when (format)
+        {
 
             "wide" -> getWideDemographicDatatable(splitData.tableData.data)
-            "long" -> splitData.tableData
+            "long", null -> splitData.tableData
```
MartinEden: Nice, I didn't know you could do that!

PR #89, opened 2017-10-23
Wide coverage data download

littledice:
Also webmodels PR: https://github.com/vimc/montagu-webmodels/pull/20


spec/MergedWideCoverageData.csvschema.json
```diff
@@ -0,0 +1,16 @@
+{
+    "columns": {
+        "scenario": "String",
+        "set_name": "String",
+        "vaccine": "String",
+        "gavi_support": ["no vaccine", "no gavi", "total"],
+        "activity_type": ["none", "routine", "campaign", "campaign-reactive"],
```
MartinEden: Could these enums be pulled out into schema files of their own? Since they are used in more than one place now.

src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/ModellingGroupController.kt
```diff
@@ -105,12 +106,69 @@ open class ModellingGroupController(context: ControllerContext)
...
+        return SplitData(splitData.structuredMetadata, tableData)
+    }
+
+    private fun getWideDatatable(data: Iterable<LongCoverageRow>):
+            FlexibleDataTable<WideCoverageRow>
+    {
+        val groupedRows = data
+                .groupBy { hashSetOf(it.countryCode, it.setName,
+                        it.ageFirst, it.ageLast,
+                        it.vaccine, it.gaviSupport, it.activityType) }//
```
MartinEden: How would you feel about laying out things like this as:

```
data.groupBy { hashSetOf(
    it.countryCode,  it.setName,
    it.ageFirst, it.ageLast,
    it.vaccine, it.gaviSupport, it.activityType
) }
```

src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/ModellingGroupController.kt
```diff
@@ -105,12 +106,69 @@ open class ModellingGroupController(context: ControllerContext)
...
+            : WideCoverageRow
+    {
+        // all records have same country, gender, age_from and age_to, so can look at first one for these
+        val reference = records.first()
+
+        val coverageAndTargetPerYear = records.associateBy(
+                { "${it.year}_coverage" },
+                { it.coverage }).plus(records.associateBy(
+                { "${it.year}_target" },
+                { it.coverage }))
```
MartinEden: Maybe easier to read as:

```
val coverageAndTargetPerYear = records.associateBy({ "${it.year}_coverage" }, { it.coverage })
                             + records.associateBy({ "${it.year}_target" }, { it.coverage })
```

MartinEden: Won't this give all the coverage columns first, then all the target columns? I think it might be easier for the modellers to use if they are interleaved? Although maybe check with Tini.

MartinEden: Wait, you've used `it.coverage` for both. The second one should be `it.target`, right?

littledice: oh yikes, good catch... clearly my unit tests are deficient! will add test and fix this

littledice: and yeah it will, I just figured this shouldn't matter, but I'll run by Tini

src/app/src/test/kotlin/org/vaccineimpact/api/tests/controllers/ModellingGroupControllersTests.kt
```diff
@@ -214,4 +304,51 @@ class ModellingGroupControllersTests : ControllerTests<ModellingGroupController>
...
+                            for (ageTo in ageTos)
+                            {
+                                for (activity in activityTypes)
+                                {
+                                    for (year in years)
+                                    {
+                                        listToReturn.add(LongCoverageRow("sId", set, vaccine, support, activity,
+                                                country, "$country-Name", year, ageFrom, ageTo, "$ageFrom-$ageTo",
+                                                random.nextDecimal(1000, 10000, numberOfDecimalPlaces = 2),
+                                                random.nextDecimal(1000, 10000, numberOfDecimalPlaces = 2)))
```
MartinEden: I prefer to avoid truly random numbers in tests, as they can lead to non-deterministic test failures that are hard to track down. Maybe have an integer outside all the for loops that goes up by one for every row, and then use the current value of that to make your coverage/target values?

src/app/src/test/kotlin/org/vaccineimpact/api/tests/controllers/ModellingGroupControllersTests.kt
```diff
@@ -118,10 +122,97 @@ class ModellingGroupControllersTests : ControllerTests<ModellingGroupController>
...
+        // there are 7 variable properties, test data includes 2 of each
+        val expectedRowCount = Math.pow(2.toDouble(), 7.toDouble())
+        Assertions.assertThat(data.count()).isEqualTo(expectedRowCount.toInt())
+
+        val expectedHeaders = listOf("1985_coverage", "1990_coverage", "1995_coverage", "2000_coverage",
+                "1985_target", "1990_target", "1995_target", "2000_target")
+
+        val firstRow = (data.first() as WideCoverageRow)
+        val values = firstRow.coverageAndTargetPerYear.values
+        val headers = firstRow.coverageAndTargetPerYear.keys
```
MartinEden: I think this is not a very clear way of checking that the long format data has been correctly transformed. Perhaps you could explicitly check the coverage and target values are as expected for one of the rows and one of the years? Obviously that would need to be in combination with the deterministic change I suggested below. Otherwise, so long as the coverage and target values are different this test will pass, even if they are e.g. the wrong way around.

src/app/src/test/kotlin/org/vaccineimpact/api/tests/controllers/ModellingGroupControllersTests.kt
littledice: You're right, this was lazy on my part. Have fixed.

littledice:
Re the format, ran by Tini and she preferred all target columns, then all coverage, but requested column headers be named like `target_[year]`, `coverage_[year]` so have amended accordingly.

spec/MergedWideCoverageData.csvschema.json
littledice: This would require some work on the CSVSchema parser. Opened a ticket for this: https://vimc.myjetbrains.com/youtrack/issue/VIMC-840

PR #92, opened 2017-10-25
Interact with triggers via function.

MartinEden:


src/config/db_version
```diff
@@ -1 +1 @@
-dc092c8
\ No newline at end of file
+i847
```
littledice: I guess change this back to latest commit?

PR #93, opened 2017-10-26
Remove deprecated fields

littledice:
* removed fields from model
* removed references to fields
* updated spec


littledice:
After this is merged we should update the api version in the webapps branch before merging https://github.com/vimc/montagu-webapps/pull/65

PR #94, opened 2017-10-31
fix duplicate demographic rows bug

littledice:
* add multiple diseases to test data
* fix bug


littledice:
When I pulled out the sql query and ran in pgAdmin the old and new queries took exactly the same amount of time. After adding multiple diseases to the large test data set, the new query seems to be very slightly quicker than just adding 'distinct' to the original select.

PR #96, opened 2017-11-02
Accept redirect url query parameter with onetime token requests

littledice:
If a request is made to `/some-endpoint/get_onetime_link/?redirectUrl=http://redirect.com`, when the resulting token is used, encode the response and redirect the user to `http://redirect.com?result=encodedResult`


littledice:
Do we want to do any validation on the redirectUrl parameter when the request to get a token is made?

src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/OneTimeLinkController.kt
```diff
@@ -28,7 +32,43 @@ class OneTimeLinkController(
...
+            if (redirectUrl.isNullOrEmpty())
+            {
+                throw e
+            }
+
+            val error = when (e)
+            {
+                is MontaguError -> e.asResult()
+                else -> UnexpectedError().asResult()
+            }
```
MartinEden: Do errors get logged if we do this? 

In general, is there a way of reusing more of the code in `ErrorHandler` to make sure the behaviour is more similar?

src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/OneTimeLinkController.kt
```diff
@@ -28,7 +32,43 @@ class OneTimeLinkController(
         val token = context.params(":token")
         val claims = verifyToken(token, repo)
         val link = OneTimeLink.parseClaims(claims)
-        return link.perform(controllers, context, repos)
+        val redirectUrl = link.queryParams["redirectUrl"]
```
MartinEden: Maybe make sure here that `redirectUrl` is always `null` or something, so that later you can use `redirectUrl == null` and smart casting.

src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/PasswordController.kt
```diff
@@ -33,6 +33,7 @@ open class PasswordController(
         return setPasswordForUser(context, repo, context.username!!)
     }
 
+    @Throws(Exception::class)
```
MartinEden: Why?

littledice: So this is a little annoying - mockito won't let you setup a method on a mock to `doThrow` unless that method explicitly does throw an exception (of the required type if you `doThrow` a specific type.) Can you think of a better way of handling this?

littledice: We could create a dummy controller with associated onetime action just for testing?

MartinEden: No, I think this is fine. Maybe just add a comment to explain why it's there?

src/app/src/main/kotlin/org/vaccineimpact/api/app/ErrorHandler.kt
```diff
@@ -26,17 +28,38 @@ open class ErrorHandler(private val serializer: Serializer = MontaguSerializer.i
         sparkException<Exception>(this::handleUnexpectedError)
```
MartinEden: Could you now get rid of this line, given that the `handleError` function can now handle non-Montagu exceptions?

src/app/src/main/kotlin/org/vaccineimpact/api/app/RedirectValidator.kt
```diff
@@ -0,0 +1,24 @@
...
+import org.vaccineimpact.api.app.errors.BadRequest
+
+open class RedirectValidator
+{
+    @Throws(BadRequest::class)
+    open fun validateRedirectUrl(redirectUrl: String?)
+    {
+        if (redirectUrl == null || redirectUrl.isEmpty()
+                || redirectUrlIsValid(redirectUrl))
+            return
```
MartinEden: Feel this is a bit unreadable.

src/app/src/main/kotlin/org/vaccineimpact/api/app/RedirectValidator.kt
```diff
@@ -0,0 +1,24 @@
...
+
+        throw BadRequest("Redirect url domain must be one of http://localhost," +
+                " https://support.montagu.dide.ic.ac.uk, https://montagu.vaccineimpact.org")
```
MartinEden: Maybe put these allowed domains in an array, even if it's not configurable, to avoid duplication below and avoid inconsistencies?

src/app/src/test/kotlin/org/vaccineimpact/api/tests/controllers/AbstractControllerTests.kt
```diff
@@ -56,6 +61,31 @@ class AbstractControllerTests : ControllerTests<AbstractController>()
...
 
     @Test
+    fun `throws error if redirect param is invalid`()
+    {
+        // Mocks
+        val parameters = mapOf(":a" to "1", ":b" to "2")
+        val context = mock<ActionContext> {
+            on { params() } doReturn parameters
+            on { username } doReturn "test.user"
+            on { queryParams("redirectUrl") } doReturn "www.redirect.com"
```
MartinEden: Maybe also check we're not allowed to redirect to a non-HTTPS version of the live site?

MartinEden: Ah, saw the tests further down. Never mind!

src/app/src/main/kotlin/org/vaccineimpact/api/app/ErrorHandler.kt
```diff
@@ -13,40 +14,50 @@ import spark.Response
...
+                        private val postgresHandler: PostgresErrorHandler = PostgresErrorHandler(),
+                        private val serializer: Serializer = MontaguSerializer.instance)
 {
-    private val logger = LoggerFactory.getLogger(ErrorHandler::class.java)
-    private val postgresHandler = PostgresErrorHandler()
+    private val unhandledExceptionMessage = "An unhandled exception occurred"
 
     init
     {
         sparkException<MontaguError>(this::handleError)
```
MartinEden: I feel like this won't work any more, as only `MontaguError`'s will get caught, not any `Exception`?

littledice: Oh yes, you're totally right, good catch

MartinEden:
This code looks good, but I won't merge because without the VPN I can't see if TeamCity has passed. I'll leave that up to you to do!

PR #97, opened 2017-11-06
moved serialization into own project

littledice:
A lot of file changes but almost all just namespace changes!


littledice:
Also this webmodels change: https://github.com/vimc/montagu-webmodels/pull/23

src/app/build.gradle
```diff
@@ -30,7 +30,6 @@ dependencies {
 
     compile "com.google.code.gson:gson:2.2.4"
```
MartinEden: Have you checked whether we can move `gson` out of this list too?

src/serialization/build.gradle
```diff
@@ -0,0 +1,20 @@
+dependencies {
+    compile "org.jetbrains.kotlin:kotlin-reflect:$kotlin_version"
+    compile 'org.pac4j:spark-pac4j:2.0.0'
+    compile 'org.pac4j:pac4j-jwt:2.0.0'
+    compile 'commons-codec:commons-codec:1.10'
+    compile 'org.abstractj.kalium:kalium:0.6.0'
+    compile "com.opencsv:opencsv:3.9"
+    compile "com.google.code.gson:gson:2.2.4"
+    compile "com.github.salomonbrys.kotson:kotson:2.5.0"
```
MartinEden: Have you confirmed that we need all of these? I'd be surprised if we needed, e.g. `kalium`.

littledice: You're right, I can thin these down.

PR #99, opened 2017-11-10
Separate out burden estimate endpoints from ModellingGroupController

MartinEden:
This is a preliminary refactor to the main work of https://vimc.myjetbrains.com/youtrack/issue/VIMC-120

I just felt like `ModellingGroupController` was getting too long.


src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/AbstractController.kt
```diff
@@ -89,15 +89,28 @@ abstract class AbstractController(controllerContext: ControllerContext,
...
         context.setResponseStatus(201)
         return url
     }
 
     fun okayResponse() = "OK"
 
-    fun buildPublicUrl(urlFragment: String) = Config["app.url"] + urlBase + urlComponent + urlFragment
+    fun buildPublicUrl(urlFragment: String, useControllerURL: Boolean = true): String
+    {
+        var url = Config["app.url"] + urlBase
```
littledice: I prefer on the whole not to use boolean parameters to change function behaviour (2 separate, clearly named, functions preferable), and also I think this hides the url building logic away from where its used. Better to just have `objectCreation` take a fully qualified url rather than`urlFragment`, and then have a helper function to build the fully qualified urls, perhaps with varags so you can pass in as many paths as you want concatenated.

Also this url building logic should be under test!

MartinEden: Good suggestions all. How about now?

PR #101, opened 2017-11-10
refactor DiseaseController

littledice:
First attempt at bringing in the alternative API architecture, where controllers are instantiated per request, routing is moved out of the controller into a RouteConfig object, and endpoints simplified.

Any thoughts? 

One thing that is a little unsatisfactory to me but I can't see a clever way around it: 

All controllers will have to have a constructor that takes an `ActionContext` and a `Repositories` object. However, ideally, for clear dependency injection and unit testing, we'd like controllers to also have a constructor in which they take the individual repos that they really want, as opposed to the whole repositories container. See the 2 constructors for `DiseaseController` to see what I mean. And in fact, this other constructor will have to be the primary one (or at least, I can't see a way of making it otherwise.) But that means that the abstract base `Controller` class can't enforce that the class has a constructor that takes a `Repositories` argument, which is a shame, because the Router depends on the existence of a constructor with that signature.

Let me know if that makes sense/if you can see a better pattern that allows us to enforce that all all controllers have a constructor with the required signature.

### DIFF
--- a/src/app/src/main/kotlin/org/vaccineimpact/api/app/app_start/Endpoint.kt
+++ b/src/app/src/main/kotlin/org/vaccineimpact/api/app/app_start/Endpoint.kt
@@ -67,11 +67,6 @@ fun Endpoint.secure(permissions: Set<String> = setOf()): Endpoint
     return this.copy(requiredPermissions = allPermissions)
 }
 
-fun Endpoint.transform(): Endpoint
-{
-    return this.copy(transform = true)
-}
-
 fun Endpoint.json(): Endpoint
 {
     return this.copy(contentType = ContentTypes.json)
--- a/src/app/src/main/kotlin/org/vaccineimpact/api/app/app_start/Endpoint.kt
+++ b/src/app/src/main/kotlin/org/vaccineimpact/api/app/app_start/Endpoint.kt
@@ -1,5 +1,6 @@
 package org.vaccineimpact.api.app.app_start
 
+import org.pac4j.sparkjava.SecurityFilter
 import org.vaccineimpact.api.app.DefaultHeadersFilter
 import org.vaccineimpact.api.app.repositories.RepositoryFactory
 import org.vaccineimpact.api.app.security.MontaguAuthorizer
@@ -52,7 +53,7 @@ data class Endpoint(
                 config,
                 configFactory.allClients(),
                 MontaguAuthorizer::class.java.simpleName,
-                "SkipOptions"
+                "method:$method"
         ))
     }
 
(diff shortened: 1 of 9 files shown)


src/app/src/main/kotlin/org/vaccineimpact/api/app/app_start/Endpoint.kt
```diff
@@ -0,0 +1,73 @@
...
+    val allPermissions = (permissions + "*/can-login").map {
+        PermissionRequirement.parse(it)
+    }
+    return this.copy(requiredPermissions = allPermissions)
+}
+
+fun Endpoint.json(): Endpoint
+{
+    return this.copy(contentType = ContentTypes.json)
+}
```
MartinEden: Is this file just a copy?

src/app/src/main/kotlin/org/vaccineimpact/api/app/app_start/EndpointDefinition.kt
```diff
@@ -0,0 +1,19 @@
...
+    val urlFragment: String
+    val controllerName: String
+    val actionName: String
+    val method: HttpMethod
+    val contentType: String
+    val transform: Boolean
+    val requiredPermissions: List<PermissionRequirement>
+
+    fun additionalSetup(url: String, webTokenHelper: WebTokenHelper, repositoryFactory: RepositoryFactory)
+}
```
MartinEden: Is this file just a copy?

MartinEden:
Would you mind creating a new branch called `i886`, instead of `ii886`, as hopefully the new make-release script will be ready soon, and it won't understand a branch name in this format.

src/app/src/main/kotlin/org/vaccineimpact/api/app/app_start/Endpoint.kt
littledice: It's almost a copy of the same file from the Reporting API... it's a substantive change to the existing Endpoint class in this project.

src/app/src/main/kotlin/org/vaccineimpact/api/app/app_start/EndpointDefinition.kt
littledice: See above comment

littledice:
One further thing that might be worth doing, after this refactor, would be to structure the top level project into 'areas' like ;Diseases', 'ModellingGroups', 'Users' etc. An area is a package containing controllers, route config, any specific webmodels needed. Something to consider.

littledice:
Closing b/c branch name is wrong. Replaced by https://github.com/vimc/montagu-api/pull/103

PR #107, opened 2017-11-15
update db interface

littledice:


littledice:
Once merged ping me on slack and I'll update the main PR - should make it more readable!

PR #116, opened 2017-11-17
I1024 onetimelinks

littledice:


littledice:
I've pulled out some test logic into a private method but only 2 use exactly the same boilerplate (others differ either in db setup, or in the assertions they make)

PR #117, opened 2017-11-20
Method to save model run info in db

littledice:
Repository work for i940. Also includes updated db interface with upload_info.uploaded_on as Timestamp.


littledice:
Also https://github.com/vimc/montagu-webmodels/pull/26

littledice:
Refactored tests to use the new structure. As discussed I'll add some validation at the deserialization level on a new branch.

MartinEden:
One more thing: You also need to add the new test class to the suite.

PR #119, opened 2017-11-20
Return estimate set type from API

MartinEden:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-1054

Once https://github.com/vimc/montagu-db/pull/38 has been merged it, `db_version` should be returned to master.

Also, review and merge #120 first.


src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/MappingHelpers.kt
```diff
@@ -0,0 +1,32 @@
...
+            deserializer.parseEnum(raw)
+        }
+        catch (e: UnknownEnumValue)
+        {
+            throw BadDatabaseConstant(e.name, e.type)
+        }
+    }
+
+    fun mapBurdenEstimateSetType(record: Record) =
+            mapBurdenEstimateSetType(
```
littledice: I like the move of mapping logic into its own class, as the repos are getting very logic heavy as we've discussed before. I'd be pro moving similar mapping logic for other types out of repos. Bu these methods are v specific to burden estimates and I think if we go with this approach we should have separate mapping logic classes, which individual repos can use as needed, rather than one giant mapping helper class. What do you think?

MartinEden: I've made changes as you suggested.

PR #120, opened 2017-11-20
Reorganize tests.

MartinEden:
This is a preliminary refactor for https://vimc.myjetbrains.com/youtrack/issue/VIMC-1054. This needs to be merged in before reviewing #119.

### DIFF
--- a/src/databaseTests/src/test/kotlin/org/vaccineimpact/api/databaseTests/RepositoryTestSuite.kt
+++ b/src/databaseTests/src/test/kotlin/org/vaccineimpact/api/databaseTests/RepositoryTestSuite.kt
@@ -5,6 +5,8 @@ import org.junit.BeforeClass
 import org.junit.runner.RunWith
 import org.junit.runners.Suite
 import org.vaccineimpact.api.databaseTests.tests.*
+import org.vaccineimpact.api.databaseTests.tests.burdenEstimateRepository.UploadBurdenEstimateTests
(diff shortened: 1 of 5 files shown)


src/databaseTests/src/test/kotlin/org/vaccineimpact/api/databaseTests/RepositoryTestSuite.kt
```diff
@@ -15,7 +17,7 @@ import org.vaccineimpact.api.test_helpers.DatabaseCreationHelper
 // Keep these sorted alphabetically, for consistency
 @RunWith(Suite::class)
 @Suite.SuiteClasses(
-        BurdenEstimateRepositoryTests::class,
+        UploadBurdenEstimateTests::class,
         BurdenEstimateRepositoryTestsDeprecated::class,
```
littledice: You'll want to add Retrieve... tests in here too

MartinEden: Fixed - good catch :)

MartinEden:
I think you're probably looking at an older branch in TeamCity. The old branch (that I ended up not using) is called `i1054_reorganize_tests`. This PR is for `i1054_refactor`.

littledice:
Ah yes, that makes more sense!

PR #123, opened 2017-11-21
method to support getting parts from multipart data

littledice:
* New method on `ActionContext` to get a named part from a multipart request.
* Tests
* Some slight refactoring of the `RequestBodySource` class in light of the above change

### DIFF
--- a/src/app/src/main/kotlin/org/vaccineimpact/api/app/context/RequestBodySource.kt
+++ b/src/app/src/main/kotlin/org/vaccineimpact/api/app/context/RequestBodySource.kt
@@ -5,31 +5,18 @@ import javax.servlet.MultipartConfigElement
 sealed class RequestBodySource
 {
     // Later, we should not return a string, but a stream or sequence of lines
-    abstract fun getBody(context: ActionContext): String
+    abstract fun getFile(context: ActionContext): String
 
     class Simple : RequestBodySource()
     {
-        override fun getBody(context: ActionContext): String
-        {
-            return context.request.body()
-        }
+        override fun getFile(context: ActionContext)
+                = context.request.body()
     }
 
-    class HTMLMultipartFile : RequestBodySource()
+    class HTMLMultipart : RequestBodySource()
     {
-        override fun getBody(context: ActionContext): String
-        {
-            val request = context.request
-            if (request.raw().getAttribute("org.eclipse.jetty.multipartConfig") == null)
-            {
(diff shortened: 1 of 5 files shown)


src/app/src/test/kotlin/org/vaccineimpact/api/tests/controllers/DirectActionContextTests.kt
```diff
@@ -85,11 +89,132 @@ class DirectActionContextTests : MontaguTests()
...
+
     private fun mockWebContext(profile: CommonProfile): SparkWebContext
     {
         val webContext = mock<SparkWebContext> {
             on { getRequestAttribute(Pac4jConstants.USER_PROFILES) } doReturn profile
         }
         return webContext
     }
+
+    private class FakePart : Part
```
MartinEden: I assume there's a reason why you can't just use a mock here?

src/app/src/main/kotlin/org/vaccineimpact/api/app/context/RequestBodySource.kt
```diff
@@ -5,31 +5,18 @@ import javax.servlet.MultipartConfigElement
 sealed class RequestBodySource
 {
     // Later, we should not return a string, but a stream or sequence of lines
-    abstract fun getBody(context: ActionContext): String
+    abstract fun getFile(context: ActionContext): String
```
MartinEden: I feel like this should remain as `getBody`, as `getFile` only makes sense in the case of the `HTMLMultipart` source. I think it would be better to, instead of harcoding "file" as the part to get, make the part name a constructor parameter of `HTMLMultipart`.

Then you can use it like `source = HTMLMultipart("file")`.

src/app/src/test/kotlin/org/vaccineimpact/api/tests/controllers/DirectActionContextTests.kt
littledice: Huh, looks to me now like absolutely no reason! Not sure what I was thinking there...

src/app/src/main/kotlin/org/vaccineimpact/api/app/context/RequestBodySource.kt
littledice: I think the opposite! `getBody` only makes sense in the `Simple` case...

PR #129, opened 2017-11-23
new endpoint to get model run parameters added to spec

littledice:


spec/spec.md
```diff
@@ -1015,13 +1032,6 @@ Schema: [`BurdenEstimates.schema.json`](BurdenEstimates.schema.json)
         }
     ]
 
-### Query parameters
-#### scenario
-Filter by scenario. e.g. GET /touchstone/2017-op-1/estimates/?scenario=menA-novacc
-
-#### responsible_group
-Filter by responsible group. e.g. GET /touchstone/2017-op-1/estimates/?responsible_group=IC-YellowFever
-
```
littledice: not part of this ticket but noticed this erroneous info

spec/spec.md
```diff
@@ -993,6 +993,23 @@ An error occurs (and no changes are made) if:
 * The coverage set is `campaign` and the scenario is `routine`. (All other combinations are acceptable)
 
 # Burden estimates
+## GET /modelling-groups/{modelling-group-id}/responsibilities/{touchstone-id}/{scenario-id}/model_run_parameter_sets/
```
MartinEden: Model run parameter sets are not scenario specific. So this URL needs to change. This may require us to change other URLs, unfortunately. This is why I tried to spec things out in advance, but I didn't manage to forsee this!

PR #131, opened 2017-11-23
Adds new endpoint with tests

littledice:
First merge https://github.com/vimc/montagu-api/pull/129 and https://github.com/vimc/montagu-api/pull/130


src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/ModellingGroupController.kt
```diff
@@ -64,6 +64,38 @@ open class ModellingGroupController(context: ControllerContext)
...
+    private fun checkValidPermissions(
+            groupId: String,
+            touchstoneId: String,
+            context: ActionContext,
+            estimateRepository: BurdenEstimateRepository,
+            readEstimatesRequired: Boolean = false
+    )
+    {
+
+        val touchstones = estimateRepository.touchstoneRepository.touchstones
```
MartinEden: Since a lot of this function is shared with `GroupBurdenEstimatesController`, maybe factor out to a common function in `CommonSecurityRules.kt`?

PR #132, opened 2017-11-27
Refactor error handling to unify logic between onetime links and normal requests

MartinEden:
This is a preliminary refactor to https://vimc.myjetbrains.com/youtrack/issue/VIMC-1088


src/app/src/main/kotlin/org/vaccineimpact/api/app/ErrorHandler.kt
```diff
@@ -32,15 +30,12 @@ open class ErrorHandler(private val logger: Logger = LoggerFactory.getLogger(Err
         val error = when (exception)
         {
             is MontaguError -> exception
-            else ->
-            {
-                logger.error(unhandledExceptionMessage, exception)
-                UnexpectedError()
-            }
+            is ValidationException -> ValidationError(exception.errors)
```
littledice: can we then remove the catch and re-throw of `ValidationException` as `ValidationError` that happens elsewhere?

littledice:
Ok, just realised this is exactly what happens in the next PR

PR #140, opened 2017-11-30
Update dependencies

MartinEden:
This updates all dependent libraries to the latest stable version. I used [this Gradle plugin](https://github.com/ben-manes/gradle-versions-plugin) to automatically detect the latest version, and then manually did find + replace.

Notes:

1. Postgres have decided to change their version number scheme, since the summer, so we have gone from version 9 to version 42. * shrug *
1. jOOQ have changed the generated code in a non-backwards compatible mode, so this branch has compilation errors. The [next PR](https://github.com/vimc/montagu-api/pull/141) then regenerates the jOOQ code - so you should review both PRs together and then merge together. But because the jOOQ code is so verbose it made this PR unreadable, so I split it out.

### DIFF
--- a/src/build.gradle
+++ b/src/build.gradle
@@ -1,6 +1,3 @@
-group 'org.vaccineimpact'
-version '1.0-SNAPSHOT'
-
 buildscript {
     ext.kotlin_version = '1.1.60'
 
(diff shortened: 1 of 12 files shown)


src/build.gradle
```diff
@@ -1,6 +1,3 @@
-group 'org.vaccineimpact'
-version '1.0-SNAPSHOT'
```
MartinEden: Plugin block must come before these, so they are moved further down

littledice:
This seems to be failing on Teamcity

PR #144, opened 2017-12-01
i1000: Refactor TouchstoneController

MartinEden:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-1000

This comes first: https://github.com/vimc/montagu-api/pull/143


src/app/src/main/kotlin/org/vaccineimpact/api/app/app_start/Endpoint.kt
```diff
@@ -17,7 +18,7 @@ data class Endpoint(
         override val actionName: String,
         override val contentType: String = ContentTypes.json,
         override val method: HttpMethod = HttpMethod.get,
-        override val transform: Boolean = false,
```
MartinEden: The `transform` property wasn't used anywhere, so I removed it while adding the new one, to keep complexity down.

src/app/src/main/kotlin/org/vaccineimpact/api/app/app_start/Router.kt
```diff
@@ -77,13 +79,33 @@ class Router(val config: RouteConfig,
         val controllerName = endpoint.controllerName
         val actionName = endpoint.actionName
 
-        val controllerType = Class.forName("org.vaccineimpact.api.app.controllers.${controllerName}Controller")
-
-        val controller = controllerType.getConstructor(ActionContext::class.java, Repositories::class.java)
-                .newInstance(context, repositories) as Controller
```
MartinEden: These two lines have been pulled out into `instantiateController`. This move may obscure the fact that there is one small change in functionality: It now looks for a three argument constructor. The new argument in `WebTokenHelper`.

src/app/src/main/kotlin/org/vaccineimpact/api/app/context/ActionContext.kt
```diff
@@ -17,6 +20,7 @@ interface ActionContext
      *  Otherwise it is null
      */
     val username: String?
+    val serializer: Serializer
```
MartinEden: Needed for `NewStyleOneTimeLinkController`.

src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/NewStyleOneTimeLinkController.kt
```diff
@@ -0,0 +1,50 @@
...
+            : this(context, repositories.token, tokenHelper)
+
+    val serializer = context.serializer
+
+    fun getTokenForDemographicData(): String
+    {
+        return getOneTimeLinkToken(context, repo, OneTimeAction.DEMOGRAPHY)
+    }
+
+    private fun getOneTimeLinkToken(
```
MartinEden: This duplicates the existing logic exactly.

src/blackboxTests/src/test/kotlin/org/vaccineimpact/api/blackboxTests/tests/ScenarioTests.kt
```diff
@@ -71,18 +71,6 @@ class ScenarioTests : DatabaseTest()
         }
     }
 
-    @Test
-    fun `only touchstone-preparer can get scenario for in-preparation touchstone`()
```
littledice: is this test obsolete for some reason?

src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/NewStyleOneTimeLinkController.kt
```diff
@@ -0,0 +1,50 @@
...
+
+class NewStyleOneTimeLinkController(
+        context: ActionContext,
+        private val repo: TokenRepository,
+        private val tokenHelper: WebTokenHelper,
+        private val redirectValidator: RedirectValidator = RedirectValidator())
+    : Controller(context)
+{
+    constructor(context: ActionContext, repositories: Repositories, tokenHelper: WebTokenHelper)
+            : this(context, repositories.token, tokenHelper)
```
littledice: No other controller is going to use the `tokenHelper`, so I'm a little reluctant for it to be a required controller parameter. (It feels ok for `Repositories` to be required because every controller needs at least one repo, and it's functioning as a dependency container to inject the right dependencies to the primary constructors.)

`KeyHelper` stores the key pair and is a singleton, so actually we're free to instantiate `WebTokenHelper` whenever right? Meaning we could pass a new instance of it to this controller without it needing to be in the base controller constructor. What do you think?

src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/endpoints/StreamedEndpoint.kt
```diff
@@ -1,16 +1,36 @@
...
+{
+    if (data is StreamSerializable<*>)
+    {
+        return stream(data, context)
+    }
+    else
+    {
+        val typeName = data?.javaClass?.simpleName ?: "unknown"
+        throw Exception("Attempted to stream $data (type: $typeName, but it is not StreamSerializable")
+    }
```
littledice: I remember this `javaClass.simpleName` property being not very human readable, see this commit and associated ticket: 
https://github.com/vimc/montagu-api/commit/e6a634f9f3ad9618c70935e56aad17d081a3b02a
https://vimc.myjetbrains.com/youtrack/issue/VIMC-744

src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/NewStyleOneTimeLinkController.kt
```diff
@@ -0,0 +1,50 @@
...
+
+    val serializer = context.serializer
```
littledice: I think it makes more sense (esp for a clean testing pattern) to leave the `ActionContext` interface as is (i.e. don't expose this property) and have this controller constructor take an instance of `Serializer`. In any case they'll be the same object, the singleton `MontaguSerializer.instance`. Just keeps dependencies clear.

src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/NewStyleOneTimeLinkController.kt
littledice: Can we have a new test to make sure this logic is covered when we remove the old method?

src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/NewStyleOneTimeLinkController.kt
MartinEden: :+1:

src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/NewStyleOneTimeLinkController.kt
MartinEden: :+1:

src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/endpoints/StreamedEndpoint.kt
MartinEden: :+1:

src/blackboxTests/src/test/kotlin/org/vaccineimpact/api/blackboxTests/tests/ScenarioTests.kt
MartinEden: Oh sorry, I meant to leave a comment about this: It was a duplicate test, presumably from a merge. There's an identical test with a slightly different name elsewhere in that class.

MartinEden:
Thanks for the feedback - very helpful 😊

littledice:
Some bizarre timestamp based test failures on TC?!

MartinEden:
It turns out we didn't have any unit tests for Router; I've added a basic one here that just proves controller instantiation and action invocation are working, so that we're not relying on the blackbox tests to catch that.

littledice:
I think this is still missing a test for the `NewStyleOneTimeLinkController`, then good to go!

MartinEden:
Sorry, completely forgot about that. Added the tests now! Good thing I did too, as it forced me to rationalize the method signature on `getOneTimeLinkToken`.

PR #145, opened 2017-12-01
i1000: Use types instead of strings

MartinEden:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-1000

First review and merge: https://github.com/vimc/montagu-api/pull/144

This isn't strictly to do with VIMC-1000, but as I was passing this seemed like an easy improvement to make. Is there any reason to prefer strings to just a reference to the class? This way if we rename the class, refactoring will take care of everything for us.


littledice:
No reason at all to be using string instead, this is something I wanted to switch over to in the Reporting API too (in fact, I'll open a ticket for it now - https://vimc.myjetbrains.com/youtrack/issue/VIMC-1120) so thanks for adding this! Can we go even further and use the actions instead of strings too?

MartinEden:
Actions seem more complicated, so I want to leave that for the moment and return to it in a future refactoring ticket.

littledice:
Cool, can you open a specific ticket for that?

PR #149, opened 2017-12-07
i949: Allow API to connect to annex via jOOQ

MartinEden:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-949

This branch doesn't compile, because it's missing the generated jOOQ code for the annex. The next PR is just the generated code, separated out for readability.


src/databaseTests/src/test/kotlin/org/vaccineimpact/api/databaseTests/tests/AnnexTests.kt
```diff
@@ -0,0 +1,22 @@
+package org.vaccineimpact.api.databaseTests.tests
```
littledice: Am I right in thinking if you remove this class for now (include in the second PR instead) the code will compile? Would be good not to have to merge in code with non passing tests!

MartinEden:
Reshuffled as requested. :)

littledice:
I'm confused! The code looks the same?

MartinEden:
How about now?

littledice:
Cool, will just wait for TC and then merge

PR #156, opened 2017-12-08
i949: Separate out logic

MartinEden:
The burden estimate writing logic is big enough that it should be its own class.


MartinEden:
There are zero logic changes here.

PR #158, opened 2017-12-11
i930: Record the current stochastic burden estimate set 

MartinEden:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-930

After https://github.com/vimc/montagu-webmodels/pull/32 is merged in, should return submodule to master.


src/databaseTests/src/test/kotlin/org/vaccineimpact/api/databaseTests/tests/burdenEstimateRepository/UploadBurdenEstimateTests.kt
```diff
@@ -0,0 +1,342 @@
+package org.vaccineimpact.api.databaseTests.tests
```
littledice: this class has been split in 2 on master - could you work out what new content was added in this branch and add it to the appropriate new class (and then remove this file?)

littledice:
There's also a TC failure after merging in master!

PR #163, opened 2017-12-12
I1128 refactor user controller

littledice:
but first merge https://github.com/vimc/montagu-api/pull/162


src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/NewStyleOneTimeLinkController.kt
```diff
@@ -14,27 +21,46 @@ import java.time.Duration
...
+        val params = mapOf(":username" to user.username)
+
+        val token = getOneTimeLinkToken(OneTimeAction.SET_PASSWORD,
+                username = user.username,
+                params = params,
+                duration = Duration.ofDays(1))
+
+        emailManager.sendEmail(NewUserEmail(user, token), user)
+        return objectCreation(context, "/users/${user.username}/")
+    }
```
MartinEden: Rather than putting this method in `NewStyleOneTimeLinkController`, maybe just factor out `getOneTimeLinkToken` into a helper class that both this controller and `UserController` new up as needed?

littledice: Yes, good shout. Having started refactoring the Password controller I've realised that this will also come in handy there.

MartinEden:
Code looks good but there's a build failure.

littledice:
Tests now passing

PR #164, opened 2017-12-12
add IP to allowed redirects

littledice:


richfitz:
Given that the IP might change it's probably worth adding a note _somewhere_ so that if it does we know when to update it here

littledice:
Yes good shout, I'll comment the code

PR #165, opened 2017-12-13
i1173: If modelling group has no responsibilities in touchstone, API returns 'null' status 

MartinEden:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-1173

First review https://github.com/vimc/montagu-webmodels/pull/34


MartinEden:
Needs pointing back at webmodels master after other PR is merged.

PR #166, opened 2017-12-14
I1128 password

littledice:
Refactoring the `PasswordController` broke the `OnetimeLink` and `OnetimeLinkController` tests, as the set password onetime action was the one used to test all the logic there.

Ended up refactoring the `OnetimeLink` class to make testing easier - rather than that data class containing the logic for invoking a callback, I created a new class `OnetimeLinkResolver` that gets passed into the controller as a dependency and so can be mocked in the `OnetimeLinkController` tests, making the setup for those tests a little clearer/simpler. 

Let me know what you think. Obviously down the line we'd like to move to onetime tokens instead of onetime links, which will simplify all this infrastructure anyway.

### DIFF
--- a/src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/MontaguControllers.kt
+++ b/src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/MontaguControllers.kt
@@ -4,12 +4,10 @@ open class MontaguControllers(context: ControllerContext)
 {
     open val modellingGroup = ModellingGroupController(context)
     open val groupBurdenEstimates = GroupBurdenEstimatesController(context)
-    open val password = PasswordController(context)
 
     val all
         get() = listOf(
                 modellingGroup,
-                groupBurdenEstimates,
-                password
+                groupBurdenEstimates
         )
 }
--- a/src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/OneTimeLinkController.kt
+++ b/src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/OneTimeLinkController.kt
@@ -3,6 +3,7 @@ package org.vaccineimpact.api.app.controllers
 import org.vaccineimpact.api.app.context.ActionContext
 import org.vaccineimpact.api.app.ErrorHandler
 import org.vaccineimpact.api.app.OneTimeLink
+import org.vaccineimpact.api.app.MontaguRedirectValidator
 import org.vaccineimpact.api.app.RedirectValidator
 import org.vaccineimpact.api.app.controllers.endpoints.oneRepoEndpoint
 import org.vaccineimpact.api.app.errors.InvalidOneTimeLinkToken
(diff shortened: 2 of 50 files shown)


src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/MontaguControllers.kt
```diff
@@ -4,12 +4,10 @@ open class MontaguControllers(context: ControllerContext)
...
     open val modellingGroup = ModellingGroupController(context)
     open val groupBurdenEstimates = GroupBurdenEstimatesController(context)
-    open val password = PasswordController(context)
 
     val all
         get() = listOf(
                 modellingGroup,
-                groupBurdenEstimates,
-                password
+                groupBurdenEstimates
```
MartinEden: Nice to see this so short!

src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/OneTimeLinkController.kt
```diff
@@ -18,9 +15,20 @@ class OneTimeLinkController(
...
         private val errorHandler: ErrorHandler = ErrorHandler(),
-        private val redirectValidator: RedirectValidator = MontaguRedirectValidator()
+        private val redirectValidator: RedirectValidator = MontaguRedirectValidator(),
+        onetimeLinkResolver: OnetimeLinkResolver? = null
 ) : AbstractController(context)
 {
+
+    private val onetimeLinkResolver = if (onetimeLinkResolver == null)
+    {
+        OnetimeLinkResolver(controllers, this.repos)
```
MartinEden: Why not just make this a default parameter value? Or failing that, you could use an elvis operator (`?:`)

src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/PasswordController.kt
```diff
@@ -58,4 +57,17 @@ open class PasswordController(
         }
         return okayResponse()
     }
+
+    private fun getSetPasswordToken(username: String): String
```
MartinEden: Shouldn't this function be moved from some existing location? But I don't see a corresponding deletion.

src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/OneTimeLinkController.kt
littledice: Can't be a default param because this.repos isn't available in the constructor. Will change to elvis though

src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/PasswordController.kt
littledice: Yes well spotted! I forgot to delete the old method from `AbstractController`, done now

PR #168, opened 2017-12-20
I1178 - New API endpoint to fetch touchstones where a modelling group has a responsibility set

georgearrowwood:
New API endpoint to fetch touchstones where a modelling group has a responsibility set


src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/ModellingGroupController.kt
```diff
@@ -109,6 +118,18 @@ open class ModellingGroupController(context: ControllerContext)
         return data.responsibilities
     }
 
+    fun getTouchstones(context: ActionContext, repo: TouchstoneRepository): List<Touchstone>
+    {
+        val groupId = groupId(context)
+
+        var touchstones = repo.getTouchstonesByGroupId(groupId)
+        if (!context.hasPermission(touchstonePreparer))
```
littledice: This logic is duplicated from the `TouchstoneController`, so would be nice to pull it out into a shared method, maybe in `CommonSecurityRules`

src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqTouchstoneRepository.kt
```diff
@@ -108,6 +108,21 @@ class JooqTouchstoneRepository(
         return ScenarioAndCoverageSets(scenario, getCoverageSetsFromRecord(records, scenario))
     }
 
+    override fun getTouchstonesByGroupId(groupId: String): List<Touchstone>
+    {
+        var query = dsl
```
littledice: Here we first need to dereference the group id, and throw an error if the group does not exist. This can all be done by the method `modellingGroupRepository.getModellingGroup(groupId)`, see the method `getModelRunParamaterSets` in the `JooqBurdenEstimateRepository` as an example of this.

We should also add a test for this (see `exception is thrown for non-existent modelling group ID` in `GetModellingGroupTests` for an example of checking for an UnknownObjectError)

src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqTouchstoneRepository.kt
```diff
@@ -288,6 +303,15 @@ class JooqTouchstoneRepository(
             mapper.mapEnum(record.status)
     )
 
+    private fun mapRecordToTouchstone(record: Record) = Touchstone(
+            record[TOUCHSTONE.ID],
```
littledice: there's already a method that does this - it's in this class, called `mapTouchstone`

src/app/src/test/kotlin/org/vaccineimpact/api/tests/controllers/ModellingGroupControllersTests.kt
```diff
@@ -364,6 +364,43 @@ class ModellingGroupControllersTests : ControllerTests<ModellingGroupController>
         controller.modifyMembership(context, mock<UserRepository>())
     }
 
+    @Test
+    fun `can return touchstones list if user has permission to read prepared touchstones`()
+    {
```
littledice: The name of these tests could be a little more descriptive - e.g. this one could be "returns in preparation touchstones if user has..." and the one below "only returns open touchstones if user has no permission...."

PR #171, opened 2017-12-22
i1227: Stream large estimates through memory (not via the disk)

MartinEden:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-1227


src/app/src/main/kotlin/org/vaccineimpact/api/app/app_start/Router.kt
```diff
@@ -67,14 +69,21 @@ class Router(val config: RouteConfig,
...
-                invokeControllerAction(endpoint, DirectActionContext(req, res), repos)
+            try
+            {
+                repositoryFactory.inTransaction { repos ->
+                    invokeControllerAction(endpoint, DirectActionContext(req, res), repos)
+                }
+            }
+            finally
+            {
+                req.consumeRemainder()
```
MartinEden: Currently we have this `try...catch` in three places, but when we finish switching over to the new style controllers that will drop down to 1.

src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/ModellingGroupController.kt
```diff
@@ -87,11 +89,13 @@ open class ModellingGroupController(context: ControllerContext)
...
-        val disease = context.getPart("disease").readText()
         context.checkEstimatePermissionsForTouchstone(groupId, touchstoneId, estimateRepository)
-        val description = context.getPart("description").readText()
 
-        val modelRuns = context.csvData<ModelRun>(RequestBodySource.HTMLMultipart("file"))
+        val parts = context.getParts()
+        val disease = parts.getPart("disease")
+        val description = parts.getPart("description")
+        val modelRuns = context.csvData<ModelRun>(parts.getPartAsRequestBodySource("file"))
+
```
littledice: This seems like quite a convoluted way to read the data from the file. How about adding an overload to `context.csvData` that accepts a string instead of a request body source? Then could just call `val modelRuns = context.csvData<ModelRun>(parts.getPart("file"))` and it would be clearer what's going on here.

src/app/src/main/kotlin/org/vaccineimpact/api/app/context/RequestBodySource.kt
```diff
@@ -17,5 +18,11 @@ sealed class RequestBodySource
         override fun getContent(context: ActionContext)
                 = context.getPart(partName)
     }
+
+    class InMemory(private val contents: String) : RequestBodySource()
+    {
```
littledice: Given that the one method this class overrides has an unused method argument, I feel like this isn't the right construction and somewhat obfuscates what's really going on. See comment below on the `ModellingGroupController` for an alternative suggestion.

src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/ModellingGroupController.kt
littledice: Also I wonder if `parts.getPart` would be better named just `parts.get`, which feels semantically the same and distinguishes it more quickly from usages of `context.getPart`

MartinEden:
Excellent suggestions; made changes as requested.

littledice:
Nice one

PR #174, opened 2017-12-22
i1217: New style controller for model run parameters

MartinEden:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-1217

Follows on from https://github.com/vimc/montagu-api/pull/172


src/app/src/main/kotlin/org/vaccineimpact/api/app/app_start/route_config/GroupModelRunParametersRouteConfig.kt
```diff
@@ -0,0 +1,33 @@
...
+                    .json()
+                    .secure(permissions),
+
+            Endpoint("$baseUrl/", controller, "addModelRunParameters")
+                    .post()
+                    .json()
+                    .secure(permissions),
+
+            Endpoint("$baseUrl/get_onetime_link/",
+                    NewStyleOneTimeLinkController::class, "getTokenForModelRunParameters")
```
littledice: We actually don't need a onetime link for this any more, as am POST-ing the file from the portal using javascript, so able to include auth headers. But maybe removing this altogether is out of scope, as this is just a refactor. I'll raise a new ticket for it.

src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/GroupModelRunParametersController.kt
```diff
@@ -0,0 +1,44 @@
...
+        return estimateRepository.getModelRunParameterSets(groupId, touchstoneId)
+    }
+
+    fun addModelRunParameters(): String
+    {
+        val touchstoneId = context.params(":touchstone-id")
+        val groupId = context.params(":group-id")
+        val disease = context.getPart("disease").readText()
+        context.checkEstimatePermissionsForTouchstone(groupId, touchstoneId, estimateRepository)
+        val description = context.getPart("description").readText()
```
littledice: I think this has reverted to the old way of consuming multipart data, rather than the updated method?

src/app/src/main/kotlin/org/vaccineimpact/api/app/app_start/Endpoint.kt
```diff
@@ -107,4 +107,9 @@ fun Endpoint.basicAuth(): Endpoint
     return this.copy(basicAuth = true)
 }
 
+fun Endpoint.post(): Endpoint
+{
+    return this.copy(method = spark.route.HttpMethod.post)
+}
```
littledice: Nice. I'm really liking this fluent API for building endpoints, feels so easy to use!

src/app/src/main/kotlin/org/vaccineimpact/api/app/app_start/route_config/GroupModelRunParametersRouteConfig.kt
littledice: Actually just realised when we refactor onetime links to use tokens, this endpoint will disappear anyway, so no need for a separate ticket

PR #180, opened 2018-01-04
I1265 - only return touchstones that have open responsibilities

littledice:
When returning touchstones for a group, only return ones where they have a responsibility set AND that set has open responsibilities in it.

### DIFF
--- a/src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqModellingGroupRepository.kt
+++ b/src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqModellingGroupRepository.kt
@@ -144,7 +144,7 @@ class JooqModellingGroupRepository(
     override fun getTouchstonesByGroupId(groupId: String): List<Touchstone>
     {
         val group = getModellingGroup(groupId)
-        var query = dsl
+        val query = dsl
                 .select(
                         TOUCHSTONE.ID,
                         TOUCHSTONE.TOUCHSTONE_NAME,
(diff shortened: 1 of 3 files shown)


src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqModellingGroupRepository.kt
```diff
@@ -144,16 +144,17 @@ class JooqModellingGroupRepository(
...
-                .select(
+        val query = dsl
+                .selectDistinct(
                         TOUCHSTONE.ID,
                         TOUCHSTONE.TOUCHSTONE_NAME,
                         TOUCHSTONE.STATUS,
                         TOUCHSTONE.DESCRIPTION,
                         TOUCHSTONE.VERSION
                 )
-                .fromJoinPath(TOUCHSTONE, RESPONSIBILITY_SET, joinType = JoinType.JOIN)
```
littledice: this joinType param not needed, as its the default value

src/blackboxTests/src/test/kotlin/org/vaccineimpact/api/blackboxTests/tests/ResponsibilityTests.kt
```diff
@@ -75,11 +75,7 @@ class ResponsibilityTests : DatabaseTest()
     fun `can get touchstones by group id`()
     {
         validate("/modelling-groups/$groupId/responsibilities/") against "Touchstones" given {
-            it.addGroup(groupId)
-            it.addTouchstoneName("touchstone", "description")
-            it.addTouchstone("touchstone", 1, description = "descr 1", status = "open")
-            it.addTouchstone("touchstone", 2)
-            it.addResponsibilitySet(groupId, touchstoneId)
```
littledice: we already have a function in this class that sets up the data as required, including open responsibilites

PR #181, opened 2018-01-08
i1261: Update API to use hyphens instead of underscores in URLs

MartinEden:
I also implemented https://vimc.myjetbrains.com/youtrack/issue/VIMC-1277 because that was the easiest way for me to check I was getting this right.


littledice:
Just needs conflict fixing

PR #182, opened 2018-01-08
Store stochastic estimates in annex

littledice:


src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqBurdenEstimateRepository.kt
```diff
@@ -226,17 +221,44 @@ class JooqBurdenEstimateRepository(
...
+
+    private fun populateStochasticBurdenEstimateSet(set: BurdenEstimateSet, disease: String,
+                                                    estimates: Sequence<BurdenEstimateWithRunId>)
+    {
+        StochasticBurdenEstimateWriter(AnnexJooqContext().dsl,
+                set.id, BurdenEstimateWriter(dsl, set.id))
+                .addEstimatesToSet(estimates, disease)
+
+        dsl.update(Tables.BURDEN_ESTIMATE_SET)
+                .set(Tables.BURDEN_ESTIMATE_SET.STATUS, "partial")
```
littledice: Setting the status to partial here rather than complete was based on the assumption that we might want to make multiple calls to populate a stochastic set from multiple files. I'm not sure if this assumption is correct?

src/databaseTests/src/test/kotlin/org/vaccineimpact/api/databaseTests/tests/burdenEstimateRepository/PopulateStochasticBurdenEstimateSetTests.kt
```diff
@@ -0,0 +1,263 @@
+package org.vaccineimpact.api.databaseTests.tests
```
littledice: Have gone with duplicating all the existing tests, just using a stochastic set instead. The stochastic estimate writer & central estimate writer are tested implicitly through these 2 test classes. I think I'm ok with this, although it would probably make for clearer tests if we refactored to make the estimate writers mockable dependenices of the repo and tested them separately...

src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqBurdenEstimateRepository.kt
MartinEden: I think it makes more sense to address partial uploads for both central and stochastic estimates in one go, when we take a look at chunked uploads (c.f. https://vimc.myjetbrains.com/youtrack/issue/VIMC-1248). For now, I think it would be better to leave the behaviour the same for both central and stochastic uploads.

src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/StochasticBurdenEstimateWriter.kt
```diff
@@ -0,0 +1,50 @@
...
+                // the other thread ("too far ahead" meaning the buffer on the input stream is full)
+                burdenEstimateWriter.writeCopyData(stream, estimates, expectedDisease, setId)
+
+                // Wait for the worker thread to finished
+                writeToDatabaseThread.join()
+            }
+        }
+    }
+
+}
```
MartinEden: I'm not sure that there is a need for two separate classes here - it feels like all the logic is the same, just a different table and fields, which could be easily parametrised, I think.

src/databaseTests/src/test/kotlin/org/vaccineimpact/api/databaseTests/tests/burdenEstimateRepository/PopulateStochasticBurdenEstimateSetTests.kt
```diff
@@ -0,0 +1,263 @@
...
+            BurdenEstimateSetType(BurdenEstimateSetTypeCode.STOCHASTIC),
+            modelRunParameterSetId = null
+    )
+
+    data class ModelRunTestData(val runParameterSetId: Int, val runs: List<Pair<String, Int>>)
+    {
+        val externalIds = runs.map { (runId, _) -> runId }
+        val internalIds = runs.map { (_, internalId) -> internalId }
+    }
+}
```
MartinEden: This seems like a lot of code to duplicate. It's generally a good idea to keep tests as simple as possible, so I can see why you've done this but... maybe we should instead have one set of tests that we run twice, once for each type? e.g. by having an abstract base class for the tests, and then two test suites that just fill in the bits that differ?

src/testHelpers/src/main/kotlin/org/vaccineimpact/api/test_helpers/InsertHelpers.kt
```diff
@@ -10,7 +10,10 @@ import org.vaccineimpact.api.models.permissions.ReifiedRole
 import org.vaccineimpact.api.security.UserHelper
 import org.vaccineimpact.api.security.ensureUserHasRole
 import java.math.BigDecimal
+import java.sql.Timestamp
+import java.time.Instant
 import java.util.*
+import kotlin.concurrent.timer
```
MartinEden: This isn't used, right?

src/testHelpers/src/main/kotlin/org/vaccineimpact/api/test_helpers/DatabaseTest.kt
```diff
@@ -9,6 +9,10 @@ abstract class DatabaseTest : MontaguTests()
 {
     private val templateDbName = Config["testdb.template_name"]
     private val dbName = Config["db.name"]
+    private val userName = Config["db.username"]
+    private val annexName = Config["annex.name"]
+    private val annexUserName = Config["annex.username"]
+    private val annexPassword = Config["annex.password"]
```
MartinEden: Are these used?

src/databaseTests/src/test/kotlin/org/vaccineimpact/api/databaseTests/tests/AnnexTests.kt
```diff
@@ -19,4 +21,15 @@ class AnnexTests : MontaguTests()
             assertThat(records).isEmpty()
         }
     }
+
+    @Test
+    fun `can read from annex`()
```
MartinEden: Might be nice to make this test name more explicit: `can read from annex via foreign table` or similar?

littledice:
Ok, have refactored the `BurdenEstimateWriter` so it can be passed in to the repo as a dependency for a cleaner testing pattern, and parameterised it to keep code DRY. This should be much clearer with less duplicated logic.

src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/BurdenEstimateWriter.kt
```diff
@@ -16,73 +18,98 @@ import java.io.PipedOutputStream
 import java.math.BigDecimal
 import kotlin.concurrent.thread
 
-class BurdenEstimateWriter(val dsl: DSLContext, val setId: Int)
+open class StochasticBurdenEstimateWriter(readDatabaseDSL: DSLContext, writeDatabaseDSL: DSLContext = AnnexJooqContext().dsl)
+    : BurdenEstimateWriter(readDatabaseDSL, writeDatabaseDSL, true)
+
+open class BurdenEstimateWriter(private val readDatabaseDSL: DSLContext,
+                                private val writeDatabaseDSL: DSLContext = readDatabaseDSL,
+                                stochastic: Boolean = false)
```
MartinEden: I like this factoring more. Nice one :)

PR #186, opened 2018-01-15
I1295 - new endpoint to set shiny cookie

littledice:


src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/AuthenticationController.kt
```diff
@@ -41,4 +41,14 @@ class AuthenticationController(context: ActionContext,
             is HTMLForm.InvalidForm -> FailedAuthentication(validationResult.problem)
         }
     }
+
+    fun setShinyCookie(): String
+    {
+        val user = userRepository.getUserByUsername(context.username!!)
+        val montaguUser = userRepository.getMontaguUserByEmail(user.email)
```
MartinEden: Why do we fetch the user from the database twice?

src/app/src/test/kotlin/org/vaccineimpact/api/tests/security/WebTokenHelperTests.kt
```diff
@@ -62,6 +62,22 @@ class WebTokenHelperTests : MontaguTests()
...
+                ReifiedRole("report.reviewer", Scope.Global())
+        )
+        val token = sut.generateShinyToken(MontaguUser(properties, roles, permissions))
+        val claims = sut.verify(token)
+
+        assertThat(claims["iss"]).isEqualTo("vaccineimpact.org")
+        assertThat(claims["sub"]).isEqualTo("test.user")
+        assertThat(claims["exp"]).isInstanceOf(Date::class.java)
+        assertThat(claims["allowed_shiny"]).isEqualTo(true)
+    }
```
MartinEden: I feel like the better place for checking that it's true when you're a report-reviewer and false otherwise is here, in the unit test, rather than the blackbox test.

MartinEden: You can leave the blackbox test if you like, now you've written it - although on the other hand, we want to keep the time spent on integration tests down, so maybe not?

src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/AuthenticationController.kt
littledice: We don't have the email initially, and don't get the roles/permissions in the first call. I could add a new repo method, I was just being lazy and re-using existing methods.

src/app/src/test/kotlin/org/vaccineimpact/api/tests/security/WebTokenHelperTests.kt
littledice: I've removed blackbox test

littledice: and added one here as suggested, you're totally right

src/app/src/test/kotlin/org/vaccineimpact/api/tests/app_start/AllowedOriginFilterTests.kt
```diff
@@ -0,0 +1,97 @@
...
+
+    @Test
+    fun `does not allow randoms`()
+    {
+        val sut = AllowedOriginsFilter(false)
+        val mockResponse = sut.handleMockRequest("https://google.com")
+        verifyZeroInteractions(mockResponse)
+    }
+    @Test
+    fun `allows localhost`()
```
MartinEden: Feel like this test should have a more descriptive name, or is just seems to contradict the earlier test `does not allow localhost`

src/app/src/test/kotlin/org/vaccineimpact/api/tests/app_start/AllowedOriginFilterTests.kt
```diff
@@ -0,0 +1,97 @@
...
+    {
+        val sut = AllowedOriginsFilter(true)
+        val mockResponse = sut.handleMockRequest("http://localhost:5000")
+        verify(mockResponse)
+                .addHeader("Access-Control-Allow-Origin", "http://localhost:5000")
+    }
+
+    @Test
+    fun `allows localhost on https`()
```
MartinEden: This test too.

src/app/src/test/kotlin/org/vaccineimpact/api/tests/app_start/AllowedOriginFilterTests.kt
```diff
@@ -0,0 +1,97 @@
...
+import com.nhaarman.mockito_kotlin.verify
+import com.nhaarman.mockito_kotlin.verifyZeroInteractions
+import org.junit.Test
+import org.vaccineimpact.api.app.app_start.AllowedOriginsFilter
+import org.vaccineimpact.api.test_helpers.MontaguTests
+import spark.Request
+import spark.Response
+import javax.servlet.http.HttpServletResponse
+
+class AllowedOriginFilterTests : MontaguTests()
```
MartinEden: It would be good to see a test that tested something other than the admin portal? Maybe just mix it up across the existing tests.

littledice:
That does sound like a better idea; I'll do it on a fresh branch for ease of reviewing.

src/app/build.gradle
```diff
@@ -21,7 +21,7 @@ if (hasProperty('docker_name')) {
 apply plugin: 'application'
 apply plugin: 'docker'
 
-mainClassName = "org.vaccineimpact.api.app.SparkAppKt"
+mainClassName = "org.vaccineimpact.api.app.app_start.SparkAppKt"
```
littledice: I moved this class because I kept expecting it to be in `app_start`

src/app/src/main/kotlin/org/vaccineimpact/api/app/MontaguRedirectValidator.kt
```diff
@@ -16,14 +16,14 @@ class MontaguRedirectValidator: RedirectValidator
...
+                Regex("^((${allowedDomains.joinToString("|") { "($it)" }}).*)")
         return validRedirectUrlPattern.matches(redirectUrl)
     }
 
     private val allowedDomains = arrayOf("http://localhost", "https://localhost",
             // NOTE this is our IP on production as of 12/12/2017
-            "https://129.31.26.29",
-            "https://support.montagu.dide.ic.ac.uk", "https://montagu.vaccineimpact.org")
+            "https://129\\.31\\.26\\.29",
+            "https://support\\.montagu\\.dide\\.ic\\.ac\\.uk", "https://montagu\\.vaccineimpact\\.org")
```
littledice: Not related to this change but I realised our redirect validator was too permissive!

src/app/src/main/kotlin/org/vaccineimpact/api/app/RequestLogger.kt
```diff
@@ -35,7 +35,7 @@ class RequestLogger(private val accessLogRepository: AccessLogRepository)
         // We can't unify the two approaches as we have different information in
         // the two cases: We have the full user retrieved from the db in the first
         // case, and only what's in the token in the second case.
-        return profile?.montaguUser()?.username
+        return profile?.internalUser()?.username
```
littledice: renamed to match the new name of the class

PR #191, opened 2018-01-16
I1112 - don't gzip if request client does not accept it

littledice:


src/app/src/main/kotlin/org/vaccineimpact/api/app/Helpers.kt
```diff
@@ -23,12 +23,18 @@ fun addTrailingSlashes(req: Request, res: Response)
...
+    = addDefaultResponseHeaders(req, res.raw(), contentType)
+
+
+fun addDefaultResponseHeaders(req: Request, res: HttpServletResponse,
+                              contentType: String = "${ContentTypes.json}; charset=utf-8")
 {
     res.contentType = contentType
-    if (res.getHeader("Content-Encoding") != "gzip")
+    val gzip = req.headers("Accept-Encoding")?.contains("gzip")
+    if (gzip != null && gzip && res.getHeader("Content-Encoding") != "gzip")
```
MartinEden: Can you do `gzip == true` instead of `gzip != null && gzip`?

littledice: Oh good shout, I wanted a `.isNullOrFalse` but didn't think of that simpler option! Doh

PR #195, opened 2018-01-17
i1321: Controller changes

MartinEden:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-1321

### DIFF
--- a/src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/GroupBurdenEstimatesController.kt
+++ b/src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/GroupBurdenEstimatesController.kt
@@ -63,6 +63,13 @@ open class GroupBurdenEstimatesController(
                 data
         )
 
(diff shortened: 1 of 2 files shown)


src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/GroupBurdenEstimatesController.kt
```diff
@@ -63,6 +63,13 @@ open class GroupBurdenEstimatesController(
                 data
         )
 
+        // Then, maybe close the burden estimate set
+        val keepOpen = context.queryParams("keepOpen")?.toBoolean() ?: true
```
littledice: so this is defaulting to true instead of false, sorry for not spotting it

littledice: oh wait, that was by design? there's a test: if keepOpen is not provided, burden estimate set is left open

PR #197, opened 2018-01-18
I1362 - refactor user repo methods

littledice:
First https://github.com/vimc/montagu-webmodels/pull/39

### DIFF
--- a/src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/UserController.kt
+++ b/src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/UserController.kt
@@ -79,15 +79,15 @@ class UserController(
         val userName = userName(context)
         val roleReadingScopes = roleReadingScopes(context)
 
-        val user = userRepository.getUserByUsername(userName)
+        val internalUser = userRepository.getUserByUsername(userName)
         if (roleReadingScopes.any())
         {
-            val roles = filteredRoles(user.roles, roleReadingScopes)
-            return user.copy(roles=roles).toUser()
+            val roles = filteredRoles(internalUser.roles, roleReadingScopes)
+            return internalUser.copy(roles=roles).toUser()
         }
         else
         {
-            return user.toUser().copy(roles = null)
+            return internalUser.toUser().copy(roles = null)
         }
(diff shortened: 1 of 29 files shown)


src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/UserController.kt
```diff
@@ -78,16 +79,15 @@ class UserController(
...
 
-        val user = userRepository.getUserByUsername(userName)
+        val internalUser = userRepository.getUserByUsername(userName)
         if (roleReadingScopes.any())
         {
-            val allRoles = userRepository.getRolesForUser(userName)
-            val roles = filteredRoles(allRoles, roleReadingScopes)
-            return user.copy(roles = roles)
+            val roles = filteredRoles(internalUser.roles, roleReadingScopes)
+            return internalUser.copy(roles=roles).toUser()
```
MartinEden: Why do you here copy and then convert to `User`, whereas in the else branch you convert to `User` and then set the roles?

src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqUserRepository.kt
```diff
@@ -1,7 +1,6 @@
 package org.vaccineimpact.api.app.repositories.jooq
```
MartinEden: There's still a chunk of duplicated logic between the two `getUser` methods. Surely the only thing that needs to differ between them is the `WHERE` clause?

littledice: Not quite (one allows returning null, the other doesn't), but have changed `getUserByEmail...` to reuse logic from the other

src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/UserController.kt
littledice: Because the InternalUser can't have a null `roles` property, but see if you find this change clearer

PR #201, opened 2018-01-22
i1322: Isolate annex database tests

MartinEden:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-1322

We need to perform tests on the annex is a separate database each time, as with the central database. Additionally, we were not closing the annex database connection at the end of using it; this becomes obvious when you try and drop the database.

### DIFF
--- a/src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/BurdenEstimateWriter.kt
+++ b/src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/BurdenEstimateWriter.kt
@@ -20,12 +20,13 @@ import kotlin.concurrent.thread
 
 open class StochasticBurdenEstimateWriter(
         readDatabaseDSL: DSLContext,
-        writeDatabaseDSLPromise: CloseableContext
+        writeDatabaseDSLPromise: CloseableContext = ShortlivedAnnexContext()
 ) : BurdenEstimateWriter(readDatabaseDSL, writeDatabaseDSLPromise, true)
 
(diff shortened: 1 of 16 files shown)


src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/BurdenEstimateWriter.kt
```diff
@@ -70,37 +71,38 @@ open class BurdenEstimateWriter(private val readDatabaseDSL: DSLContext,
...
-                        modelRuns,
-                        modelRunParameterId,
-                        stream,
-                        estimates,
-                        expectedDisease,
-                        setId)
-
-                // Wait for the worker thread to finished
-                writeToDatabaseThread.join()
+        writeDatabaseDSLPromise.inside { writeDatabaseDSL ->
```
MartinEden: Only change here is indentation due to wrapping everything in this new call to "inside".

src/databaseInterface/src/main/kotlin/org/vaccineimpact/api/db/CloseableContext.kt
```diff
@@ -0,0 +1,33 @@
+package org.vaccineimpact.api.db
```
MartinEden: I don't think my naming is very good here. Any suggestions?

littledice: This seems ok, I can't think of anything better really!

PR #204, opened 2018-01-22
I1372 - duplicate key error not being returned

littledice:
Sometimes the informative PSQL error is nested (3 times deep in the case of the burden estimate writer) - so lets check for a duplicate key error recursively.

### DIFF
--- a/spec/ErrorCode.schema.json
+++ b/spec/ErrorCode.schema.json
@@ -2,5 +2,5 @@
     "id": "ErrorCode",
     "type": "string",
     "minLength": 1,
-    "pattern": "^[a-zA-Z0-9]+(-[a-zA-Z0-9]+)*(:[a-zA-Z0-9_]+(-[a-zA-Z0-9_]+)*)*$"
+    "pattern": "^[a-zA-Z0-9]+(-[a-zA-Z0-9]+)*(:[a-zA-Z0-9_,\\s]+(-[a-zA-Z0-9_]+)*)*$"
 }
--- a/spec/ErrorCode.schema.json
+++ b/spec/ErrorCode.schema.json
@@ -2,5 +2,5 @@
     "id": "ErrorCode",
     "type": "string",
     "minLength": 1,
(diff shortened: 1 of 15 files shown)


spec/ErrorCode.schema.json
```diff
@@ -2,5 +2,5 @@
     "id": "ErrorCode",
     "type": "string",
     "minLength": 1,
-    "pattern": "^[a-zA-Z0-9]+(-[a-zA-Z0-9]+)*(:[a-zA-Z0-9_]+(-[a-zA-Z0-9_]+)*)*$"
+    "pattern": "^[a-zA-Z0-9]+(-[a-zA-Z0-9]+)*(:[a-zA-Z0-9_,\\s]+(-[a-zA-Z0-9_]+)*)*$"
```
littledice: in the case of duplicate burden estimate rows, the error code contains a comma separated list of fields from the violated unique constraint

src/blackboxTests/src/test/kotlin/org/vaccineimpact/api/blackboxTests/tests/BurdenEstimates/PopulateBurdenEstimateTests.kt
```diff
@@ -158,4 +158,20 @@ class PopulateBurdenEstimateTests : BurdenEstimateTests()
...
+    {
+        TestUserHelper.setupTestUser()
+        val setId = JooqContext().use {
+            setUpWithBurdenEstimateSet(it)
+        }
+        val token = TestUserHelper.getToken(requiredWritePermissions, includeCanLogin = true)
+        val helper = RequestHelper()
+        val response = helper.post("$setUrl/$setId/", duplicateCsvData, token = token)
+        JSONValidator()
+                .validateError(response.text, "duplicate-key:burden_estimate_set, country, year, age, burden_outcome")
```
MartinEden: I don't mind having commas in our machine readable error codes, but the spaces are a bit unpleasant. Would it be straightforward to remove them?

littledice: Sure, we could do that in the `PostgresErrorHandler`

PR #205, opened 2018-01-22
i1322: New endpoint

MartinEden:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-1322

First review #201, #202 and #203.

### DIFF
--- a/src/blackboxTests/src/test/kotlin/org/vaccineimpact/api/blackboxTests/schemas/EmptySchema.kt
+++ b/src/blackboxTests/src/test/kotlin/org/vaccineimpact/api/blackboxTests/schemas/EmptySchema.kt
@@ -1,0 +1,32 @@
+package org.vaccineimpact.api.blackboxTests.schemas
+
+import org.assertj.core.api.Assertions.assertThat
+import org.vaccineimpact.api.validateSchema.Validator
+
+object EmptySchema : Schema
+{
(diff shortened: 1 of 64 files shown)


src/blackboxTests/src/test/kotlin/org/vaccineimpact/api/blackboxTests/schemas/EmptySchema.kt
```diff
@@ -0,0 +1,32 @@
...
+    {
+        assertThat(response).isBlank()
+    }
+
+    override fun validateRequest(request: String)
+    {
+        assertThat(request).isBlank()
+    }
+
+    class EmptyValidator : Validator
```
littledice: This is never used right? Why not just set the `validator` to be an instance of the existing `JsonValidator`? Or throw an exception on invocation like we do with the `CSVSchema` (as presumably we don't ever want a blank response?)

littledice: Unless I'm missing something!

MartinEden: Yeah, you're right - I was just blindly implementing the interface. Fixed.

PR #207, opened 2018-01-23
I1369 - clear cookie endpoint

littledice:


src/app/src/main/kotlin/org/vaccineimpact/api/app/app_start/route_config/AuthenticationRouteConfig.kt
```diff
@@ -14,6 +14,8 @@ object AuthenticationRouteConfig : RouteConfig
                     .basicAuth(),
             Endpoint("/set-shiny-cookie/", controller, "setShinyCookie")
                     .secure()
+                    .json(),
+            Endpoint("/clear-shiny-cookie/", controller, "clearShinyCookie")
```
MartinEden: I think you're right that this doesn't need to be `secure`, but given that it's a bit unusual in the API, maybe add a comment here explaining why, so it doesn't surprise/confuse us in the future?

src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/AuthenticationController.kt
```diff
@@ -50,4 +50,11 @@ class AuthenticationController(context: ActionContext,
         context.addResponseHeader("Access-Control-Allow-Credentials", "true")
         return okayResponse()
     }
+
+    fun clearShinyCookie(): String
+    {
+        context.addResponseHeader("Set-Cookie", "jwt_token=; Path=/; Secure; HttpOnly; SameSite=Lax")
```
MartinEden: Maybe these lines could be factored out to be shared with `setShinyCookie`, and just pass an empty token string for the clear case?

src/app/src/main/kotlin/org/vaccineimpact/api/app/app_start/route_config/AuthenticationRouteConfig.kt
littledice: yes good idea

PR #208, opened 2018-01-23
I1363 - check set belongs to responsibility

littledice:
- when populating a set
- when clearing a set


littledice:
Right. `getResponsibilityInfo` gets the responsibility for that group, for that touchstone, for that scneario.

PR #209, opened 2018-01-24
I1272 - order responsibilities sensibly

littledice:
The reported issue was that the responsibilities here are not ordered: https://support.montagu.dide.ic.ac.uk:10443/contribution/VIMC/responsibilities/201801rfp-1/

The request was that they be ordered 'No Vacc, Routine, Campaign', i.e. in the order of the API's `ActivityType` enum, but these scenario descriptions can be any string (i.e. won't neccessarily map nicely onto the enum values), while activity type is only stored against coverage set in the database. So I'm joining to the focal coverage set when retrieving scenarios, and ordering by the activity type of that focal coverage set.

### DIFF
--- a/src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqScenarioRepository.kt
+++ b/src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqScenarioRepository.kt
@@ -4,13 +4,16 @@ import org.jooq.DSLContext
 import org.jooq.Record
 import org.vaccineimpact.api.app.errors.UnknownObjectError
 import org.vaccineimpact.api.app.repositories.ScenarioRepository
-import org.vaccineimpact.api.db.Tables.SCENARIO
-import org.vaccineimpact.api.db.Tables.SCENARIO_DESCRIPTION
+import org.vaccineimpact.api.app.repositories.jooq.mapping.MappingHelper
+import org.vaccineimpact.api.db.Tables.*
 import org.vaccineimpact.api.db.fieldsAsList
 import org.vaccineimpact.api.db.fromJoinPath
+import org.vaccineimpact.api.models.ActivityType
 import org.vaccineimpact.api.models.Scenario
 
-class JooqScenarioRepository(dsl: DSLContext) : JooqRepository(dsl), ScenarioRepository
+class JooqScenarioRepository(dsl: DSLContext,
+                             private val mapper: MappingHelper = MappingHelper())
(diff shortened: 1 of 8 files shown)


src/databaseTests/src/test/kotlin/org/vaccineimpact/api/databaseTests/tests/touchstoneRepository/GetScenarioTests.kt
```diff
@@ -90,9 +90,11 @@ class GetScenarioTests : TouchstoneRepositoryTests()
 
     private fun checkScenarioIsAsExpected(result: ScenarioAndCoverageSets, extraTouchstones: List<String> = emptyList())
     {
-        assertThat(result.scenario).isEqualTo(Scenario(
-                "yf-1", "Yellow Fever 1", "YF", listOf(touchstoneId) + extraTouchstones
-        ))
+        assertThat(result.scenario.description).isEqualTo("Yellow Fever 1")
+        assertThat(result.scenario.id).isEqualTo("yf-1")
+        assertThat(result.scenario.disease).isEqualTo("YF")
+        assertThat(result.scenario.touchstones).hasSameElementsAs(listOf(touchstoneId) + extraTouchstones)
```
littledice: Making this test invariant under ordering of the touchstones, as I don't think we care about this (ordering by touchstone id seems reasonable, but happens to flip the order in some tests here)

src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqScenarioRepository.kt
```diff
@@ -23,14 +26,38 @@ class JooqScenarioRepository(dsl: DSLContext) : JooqRepository(dsl), ScenarioRep
...
+        val scenario = mapScenario(input)
+        val activityType = first[COVERAGE_SET.ACTIVITY_TYPE]
+        val activityTypeEnum = if (activityType != null)
+        {
+            mapper.mapEnum<ActivityType>(activityType)
+        }
+        else
+        {
+            null
+        }
```
MartinEden: One small readability suggestion: We could add a method `mapNullableEnum` method to mapper, that handles this null check for us. Then this code could be the more readable: 
```
val activityType = mapper.mapNullableEnum(first[COVERAGE_SET.ACTIVITY_TYPE])
```

littledice: Nice idea

PR #215, opened 2018-01-26
i1398: Proper logic around model run parameter sets

MartinEden:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-1398


MartinEden:
This turned out more complex than I hoped for two reasons:
* I had to introduce a new, more complex kind of validation rule in order to be able to validate one field based on the value of another.
* I had to update `FluentValidation` so we could get the ID of the burden estimate set out in order to do further tests on it.

src/app/src/main/kotlin/org/vaccineimpact/api/app/models/CreateUser.kt
```diff
@@ -1,7 +1,7 @@
 package org.vaccineimpact.api.app.models
 
+import org.vaccineimpact.api.models.validation.AllowedFormat
 import org.vaccineimpact.api.security.BasicUserProperties
-import org.vaccineimpact.api.serialization.validation.AllowedFormat
```
MartinEden: I've moved the validation annotations into the submodule so that we can use them on models.

src/serialization/src/main/kotlin/org/vaccineimpact/api/serialization/ModelBinder.kt
```diff
@@ -33,15 +34,12 @@ class ModelBinder(private val serializer: Serializer = MontaguSerializer.instanc
         @Suppress("UNCHECKED_CAST")
         val klass = model::class as KClass<Any>
         val errors = mutableListOf<ErrorInfo>()
-        val name = property.name
+        val name = serializer.convertFieldName(property.name)
```
MartinEden: Up until now all the fields have been single words, like `username`, so we hadn't been needing to convert them to snake_case.

src/serialization/src/main/kotlin/org/vaccineimpact/api/serialization/ModelBinder.kt
```diff
@@ -33,15 +34,12 @@ class ModelBinder(private val serializer: Serializer = MontaguSerializer.instanc
...
         val value = property.get(model)
 
-        if (value == null && !property.returnType.isMarkedNullable)
+        if (!property.returnType.isMarkedNullable)
         {
-            errors += ErrorInfo(
-                    "invalid-field:$name:missing",
-                    "You have not supplied a value, or have supplied a null value, for field '$name'"
-            )
+            errors += listOfNotNull(checkMissing(value, name))
```
MartinEden: This change is so we can share the "error on null" logic with the new `RequiredWhen` annotation.

src/serialization/src/main/kotlin/org/vaccineimpact/api/serialization/validation/ValidationRules.kt
```diff
@@ -1,16 +1,17 @@
...
 
 import org.vaccineimpact.api.models.ErrorInfo
+import org.vaccineimpact.api.models.validation.AllowedFormat
+import org.vaccineimpact.api.models.validation.MinimumLength
+import org.vaccineimpact.api.models.validation.RequiredWhen
+import kotlin.reflect.full.memberFunctions
 
-annotation class CanBeBlank
-annotation class AllowedFormat(val pattern: String, val example: String)
-annotation class MinimumLength(val length: Int)
```
MartinEden: The annotations have moved into the submodule

PR #220, opened 2018-01-29
cookie is only secure if allowlocalhost is false

littledice:
Allow the shiny cookie to be set over an insecure connection when `allowLocalhost` is set to true.


MartinEden:
Needs conflicts fixing

PR #222, opened 2018-01-31
I1419 - browser hangs with long redirect url

littledice:
Solution is to re-throw errors rather than encoding them and redirecting if the encoded result is longer than 2000 chars.

I'm quite puzzled because my best googling can't turn up any stack overflow questions, browser specs, or any form of documentation about this phenomenon. From testing found that the breakpoint (with chrome) is somewhere around the 8000 character mark, but chose 1900 because I did find that IE has a url length limit, and keeping the query string under 1900 characters will keep us just within that limit: https://support.microsoft.com/en-us/help/208427/maximum-url-length-is-2-083-characters-in-internet-explorer

As I write this, it occurs to me that I don't think it is a browser issue at all. I think it is the Spark server failing to issue the redirect. I'm going to try debugging a little further, but either way seems like it's a limitation we probably can't (and shouldn't need to) overcome.

Further subtasks to prevent a request ever getting to this stage, i.e. checking that the POST-ed file is text and not binary data.

### DIFF
--- a/src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/OneTimeLinkController.kt
+++ b/src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/OneTimeLinkController.kt
@@ -98,6 +98,8 @@ class OneTimeLinkController(
     {
         val encodedResult = tokenHelper.encodeResult(result)
         context.request.consumeRemainder()
+        // it is recommended to keep urls under 2000 characters
+        // https://stackoverflow.com/questions/417142/what-is-the-maximum-length-of-a-url-in-different-browsers/417184#417184
         if (encodedResult.length > 1900 && exception != null)
             throw exception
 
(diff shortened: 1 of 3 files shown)


src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/OneTimeLinkController.kt
```diff
@@ -91,10 +93,14 @@ class OneTimeLinkController(
         return oneTimeTokenGenerator.getOneTimeLinkToken(OneTimeAction.BURDENS_POPULATE, context)
     }
 
-    private fun redirectWithResult(context: ActionContext, result: Result, redirectUrl: String)
+    private fun redirectWithResult(context: ActionContext, result: Result, redirectUrl: String,
+                                   exception: Exception? = null)
     {
         val encodedResult = tokenHelper.encodeResult(result)
         context.request.consumeRemainder()
+        if (encodedResult.length > 1900 && exception != null)
```
MartinEden: Maybe a comment here explaining why?

littledice:
I think I finally found something: 

https://stackoverflow.com/questions/31792358/programmatically-set-jetty-configuration-to-increase-allowed-url-length

MartinEden:
Even so, there will still be some upper limit, so your changes to handle that situation gracefully are worth having.

littledice:
Yep, I think 2000 chars is the right limit, based on what I've read here: https://boutell.com/newfaq/misc/urllength.html

PR #224, opened 2018-02-01
i1433: Fix hanging test

MartinEden:
I think I've got it right this time!


MartinEden:
The problem was that the write to database thread waits for the main thread to write the terminator `\.`. We do this when we leave the `PostgresCopyWriter`'s `use` block. However, this test was triggering an error in the one line of code that wasn't inside the block, so we never entered the block, and so we never left it, and never wrote anything to the stream at all.

PR #225, opened 2018-02-01
i1405: Bring over code that generates onetime tokens

MartinEden:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-1405

This adds a new endpoint for generating onetime tokens in the same way as the reporting API. Currently this is not useful for anything, as there isn't a corresponding security filter.

### DIFF
--- a/src/app/src/main/kotlin/org/vaccineimpact/api/app/app_start/route_config/OneTimeLinkRouteConfig.kt
+++ b/src/app/src/main/kotlin/org/vaccineimpact/api/app/app_start/route_config/OneTimeLinkRouteConfig.kt
@@ -1,7 +1,6 @@
 package org.vaccineimpact.api.app.app_start.route_config
 
 import org.vaccineimpact.api.app.app_start.Endpoint
-import org.vaccineimpact.api.app.app_start.json
 import org.vaccineimpact.api.app.app_start.secure
 import org.vaccineimpact.api.app.controllers.OneTimeLinkController
 import spark.route.HttpMethod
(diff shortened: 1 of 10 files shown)


src/security/src/test/kotlin/SodiumTests.kt
```diff
@@ -1,8 +1,9 @@
 import org.assertj.core.api.Assertions.assertThat
 import org.junit.Test
 import org.vaccineimpact.api.security.SodiumPasswordEncoder
+import org.vaccineimpact.api.test_helpers.MontaguTests
 
-class SodiumTests
+class SodiumTests : MontaguTests()
```
MartinEden: Just noticed in passing that these didn't get the TeamCity messages

src/app/src/main/kotlin/org/vaccineimpact/api/app/app_start/route_config/OneTimeLinkRouteConfig.kt
```diff
@@ -10,6 +11,11 @@ object OneTimeLinkRouteConfig : RouteConfig
     val url = "/onetime_link/:token/"
 
     override val endpoints = listOf(
+            // This gets new style onetime tokens (like the reporting API)
+            Endpoint("/onetime_token/", controller, "getToken")
```
littledice: didn't we at some point move over all underscores in endpoints to hyphens?

MartinEden: As discussed, I've opened https://vimc.myjetbrains.com/youtrack/issue/VIMC-1440 to do this properly.

PR #229, opened 2018-02-06
I1470 - return past touchstones with closed responsibilities

littledice:
We only want groups to see open touchstones in which they have open responsibilities. But they should be able to see past touchstones in which they have closed responsibilities. It's also fine to return in-prep touchstones with closed responsibilities.


MartinEden:
Does filtering out `in-preparation` touchstones (if you don't have permission to see them) happen at the controller level? If so, this looks fine and I'm happy for you to merge.

littledice:
Yes, exactly, that filtering happen in the controller

PR #230, opened 2018-02-07
I1426 - Server error when RequiredWhen depends on null value

littledice:
A couple of changes.

1. Check for null non-nullable properties recursively, so for e.g. `{ type: { } }` as well as `{}` throws a validation exception when creating a burden estimate set
2. Add further varargs to the `RequiredWhen` attribute which indicate which properties of the model the required when logic depends on. These are then checked for invalid nulls first, and if invalid the function is never called. This does mean the null check is run twice on any such fields; we could do something more complicated like maintain a dictionary of field names and validity statuses so we can just look them up the second time...

### DIFF
--- a/src/app/src/test/kotlin/org/vaccineimpact/api/tests/models/CreateBurdenEstimateSetTests.kt
+++ b/src/app/src/test/kotlin/org/vaccineimpact/api/tests/models/CreateBurdenEstimateSetTests.kt
@@ -29,7 +29,6 @@ class CreateBurdenEstimateSetTests : ValidationTests()
         }
     }
 
-
     @Test
     fun `missing type throws validation exception`()
     {
--- a/src/app/src/test/kotlin/org/vaccineimpact/api/tests/models/CreateBurdenEstimateSetTests.kt
+++ b/src/app/src/test/kotlin/org/vaccineimpact/api/tests/models/CreateBurdenEstimateSetTests.kt
@@ -25,7 +25,7 @@ class CreateBurdenEstimateSetTests : ValidationTests()
         {
             val errors = e.errors
             Assertions.assertThat(errors.count()).isEqualTo(1)
-            Assertions.assertThat(errors[0].code).isEqualTo("invalid-field:type:missing")
+            Assertions.assertThat(errors[0].code).isEqualTo("invalid-field:create_burden_estimate_set:type:missing")
         }
     }
 
@@ -45,7 +45,7 @@ class CreateBurdenEstimateSetTests : ValidationTests()
         {
             val errors = e.errors
             Assertions.assertThat(errors.count()).isEqualTo(1)
-            Assertions.assertThat(errors[0].code).isEqualTo("invalid-field:type:missing")
+            Assertions.assertThat(errors[0].code).isEqualTo("invalid-field:create_burden_estimate_set:type:missing")
         }
     }
 
@@ -65,7 +65,7 @@ class CreateBurdenEstimateSetTests : ValidationTests()
         {
             val errors = e.errors
             Assertions.assertThat(errors.count()).isEqualTo(1)
-            Assertions.assertThat(errors[0].code).isEqualTo("invalid-field:type:missing")
+            Assertions.assertThat(errors[0].code).isEqualTo("invalid-field:burden_estimate_set_type:type:missing")
         }
     }
 
@@ -93,7 +93,7 @@ class CreateBurdenEstimateSetTests : ValidationTests()
         {
             val errors = e.errors
             Assertions.assertThat(errors.count()).isEqualTo(1)
-            Assertions.assertThat(errors[0].code).isEqualTo("invalid-field:model_run_parameter_set:missing")
+            Assertions.assertThat(errors[0].code).isEqualTo("invalid-field:create_burden_estimate_set:model_run_parameter_set:missing")
         }
     }
 
--- a/src/app/src/test/kotlin/org/vaccineimpact/api/tests/models/CreateBurdenEstimateSetTests.kt
+++ b/src/app/src/test/kotlin/org/vaccineimpact/api/tests/models/CreateBurdenEstimateSetTests.kt
@@ -2,8 +2,6 @@ package org.vaccineimpact.api.tests.models
 
 import org.assertj.core.api.Assertions
 import org.junit.Test
-import org.vaccineimpact.api.models.BurdenEstimateSetType
-import org.vaccineimpact.api.models.BurdenEstimateSetTypeCode
 import org.vaccineimpact.api.models.CreateBurdenEstimateSet
 import org.vaccineimpact.api.serialization.validation.ValidationException
 
@@ -15,17 +13,11 @@ class CreateBurdenEstimateSetTests : ValidationTests()
         val badModelJson = "{}"
         Assertions.assertThatThrownBy {
             binder.deserialize<CreateBurdenEstimateSet>(badModelJson, CreateBurdenEstimateSet::class.java)
-        }
-
-        try
-        {
-            binder.deserialize<CreateBurdenEstimateSet>(badModelJson, CreateBurdenEstimateSet::class.java)
-        }
-        catch (e: ValidationException)
-        {
-            val errors = e.errors
-            Assertions.assertThat(errors.count()).isEqualTo(1)
+        }.matches {
+            val errors = (it as ValidationException).errors
+            Assertions.assertThat(errors).hasSize(1)
             Assertions.assertThat(errors[0].code).isEqualTo("invalid-field:create_burden_estimate_set:type:missing")
+            true
         }
     }
 
@@ -35,17 +27,11 @@ class CreateBurdenEstimateSetTests : ValidationTests()
         val badModelJson = "{ \"modelRunParameterSet\" : \"1\"}"
         Assertions.assertThatThrownBy {
             binder.deserialize<CreateBurdenEstimateSet>(badModelJson, CreateBurdenEstimateSet::class.java)
-        }
-
-        try
-        {
-            binder.deserialize<CreateBurdenEstimateSet>(badModelJson, CreateBurdenEstimateSet::class.java)
-        }
-        catch (e: ValidationException)
-        {
-            val errors = e.errors
-            Assertions.assertThat(errors.count()).isEqualTo(1)
+        }.matches {
+            val errors = (it as ValidationException).errors
+            Assertions.assertThat(errors).hasSize(1)
             Assertions.assertThat(errors[0].code).isEqualTo("invalid-field:create_burden_estimate_set:type:missing")
+            true
         }
     }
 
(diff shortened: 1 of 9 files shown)


src/app/src/test/kotlin/org/vaccineimpact/api/tests/models/CreateUserTests.kt
```diff
@@ -19,27 +19,27 @@ class CreateUserTests : ValidationTests()
...
-        assertCausesTheseErrors(model.copy(username = "user_name"), "invalid-field:username:bad-format")
-        assertCausesTheseErrors(model.copy(username = "12345"), "invalid-field:username:bad-format")
-        assertCausesTheseErrors(model.copy(username = "name&"), "invalid-field:username:bad-format")
+        assertCausesTheseErrors(model.copy(username = ""), "invalid-field:create_user:username:blank")
+        assertCausesTheseErrors(model.copy(username = "user name"), "invalid-field:create_user:username:bad-format")
+        assertCausesTheseErrors(model.copy(username = "User.Name"), "invalid-field:create_user:username:bad-format")
+        assertCausesTheseErrors(model.copy(username = "user-name"), "invalid-field:create_user:username:bad-format")
+        assertCausesTheseErrors(model.copy(username = "user_name"), "invalid-field:create_user:username:bad-format")
+        assertCausesTheseErrors(model.copy(username = "12345"), "invalid-field:create_user:username:bad-format")
+        assertCausesTheseErrors(model.copy(username = "name&"), "invalid-field:create_user:username:bad-format")
```
littledice: I made the field name more qualified in the error codes, to distinguish `create_burden_estimate_set:type` from `burden_estimate_type_code: type` but it does make these ones perhaps less clear? Maybe we could qualify them only for nested properties.

src/serialization/src/main/kotlin/org/vaccineimpact/api/serialization/ModelBinder.kt
```diff
@@ -3,29 +3,30 @@ package org.vaccineimpact.api.serialization
...
+                  private val validator: Validator = Validator(serializer))
 {
     fun <T : Any> deserialize(body: String, klass: Class<T>): T
     {
         val model = serializer.fromJson(preprocessed(body), klass)
         val errors = verify(model)
         if (errors.any())
         {
-            throw ValidationException(errors)
+            throw ValidationException(errors.distinct())
```
littledice: A `RequiredWhen` check may return a validation error for a null dependent property, which will also be returned by the regular null check, so there could be duplicates.

src/app/src/test/kotlin/org/vaccineimpact/api/tests/models/CreateBurdenEstimateSetTests.kt
```diff
@@ -0,0 +1,100 @@
...
+        {
+            binder.deserialize<CreateBurdenEstimateSet>(badModelJson, CreateBurdenEstimateSet::class.java)
+        }
+        catch (e: ValidationException)
+        {
+            val errors = e.errors
+            Assertions.assertThat(errors.count()).isEqualTo(1)
+            Assertions.assertThat(errors[0].code).isEqualTo("invalid-field:create_burden_estimate_set:type:missing")
+        }
+    }
```
MartinEden: I think the test structure you are looking for here is:

```
val badModelJson = "{}"
Assertions.assertThatThrownBy {
    binder.deserialize<CreateBurdenEstimateSet>(badModelJson, CreateBurdenEstimateSet::class.java)
}
        .isInstanceOf(ValidationException::class.java)
        .matches {
            val errors = (it as ValidationException).errors
            assertThat(errors).hasSize(1)
            assertThat(errors[0].code).isEqualTo("invalid-field:create_burden_estimate_set:type:missing")
            true
        }
```

Also, it might be good to factor this out, so you can write the tests like so:

```
assertThrowsThisMontaguError("invalid-field:create_burden_estimate_set:type:missing") {
      binder.deserialize(badModelJson, CreateBurdenEstimateSet::class.java)
}
```

src/app/src/test/kotlin/org/vaccineimpact/api/tests/models/CreateUserTests.kt
MartinEden: Do you need the top level object? Couldn't it just be `invalid-field:type:type:missing` and `invalid-field:username:blank`

littledice:
Full recursive validation seems like a good shout, rather than special casing, will amend. You can't solve the burden estimate bug the way you suggest precisely because the type parameter is non-nullable, so the kotlin compiler won't allow a null check. We could make the type field nullable, but that would be a mistake I think!

littledice:
Have added full recursive validation but it alone won't fix the `RequiredWhen` bug.

For properties required to determine whether the `RequiredWhen` condition is satisfied, we will need to either to store validation results in a lookup we can read from, or make a further recursive null check before executing the `RequiredWhen` function...

littledice:
Or only validate `RequiredWhen` attributes when all other validation rules have passed. But I think it's nice to get all errors at once, and this would complicate the code a fair bit.

littledice:
Actually you're right, the null check on a non-nullable object is just a warning, not a compilation error. So we could just do that and suppress the warning.

src/serialization/src/main/kotlin/org/vaccineimpact/api/serialization/ModelBinder.kt
```diff
@@ -3,53 +3,75 @@ package org.vaccineimpact.api.serialization
...
         }
         return model
     }
 
     fun verify(model: Any): List<ErrorInfo>
     {
         val properties = model::class.memberProperties.filterIsInstance<KProperty1<Any, *>>()
-        return properties.flatMap { verify(it, model) }
+        return properties.flatMap { recursiveVerify(it, model) ?: listOf() }
```
MartinEden: Can `recursiveVerify` return null? If not, why do you need an elvis operator here?

src/serialization/src/main/kotlin/org/vaccineimpact/api/serialization/ModelBinder.kt
```diff
@@ -3,53 +3,75 @@ package org.vaccineimpact.api.serialization
...
+        if (!klass.isData)
+        {
+            // end condition
+            // primitive property
+            return errors
+        }
+
+        val members = klass.memberProperties.filterIsInstance<KProperty1<Any, *>>()
+        return errors + members.flatMap { recursiveVerify(it, value) }
+    }
```
MartinEden: Maybe clearer just to write this with a mutable errors? Then you can have single point of return. i.e.

```
var errors = verify(property, model)
val value = property.get(model)
if (value != null && value::class.isData)
{
    val members = value::class.memberProperties.filterIsInstance<KProperty1<Any, *>>()
    errors += members.flatMap { recursiveVerify(it, value) }
}
return errors
```

src/serialization/src/main/kotlin/org/vaccineimpact/api/serialization/ModelBinder.kt
```diff
@@ -3,53 +3,75 @@ package org.vaccineimpact.api.serialization
...
-            errors += ErrorInfo(
-                    "invalid-field:$name:blank",
-                    "You have supplied an empty or blank string for field '$name'"
-            )
+            return errors
         }
-        errors += property.allAnnotations(klass).map { applyRule(it, value, name, model) }.filterNotNull()
-        return errors
+
+        return listOf()
```
MartinEden: Why the "if any, otherwise empty" logic? Why not just `return errors`?

src/serialization/src/main/kotlin/org/vaccineimpact/api/serialization/ModelBinder.kt
```diff
@@ -3,53 +3,75 @@ package org.vaccineimpact.api.serialization
...
+        errors += validator.nullCheck(property.get(model), property, model, name)
+
+        if (property.findAnnotationAnywhere<CanBeBlank>(klass) == null)
         {
-            errors += listOfNotNull(checkMissing(value, name))
+            errors += validator.checkBlank(property, name, model)
         }
-        if (value is String && value.isBlank() && property.findAnnotationAnywhere<CanBeBlank>(klass) == null)
+
+        errors += property.allAnnotations(klass).flatMap { validator.applyRule(it, property, name, model) }
```
MartinEden: Why have you switched to passing in the property rather than the value here? Presumably it's a bit slower to keep getting the value by reflection?

src/serialization/src/main/kotlin/org/vaccineimpact/api/serialization/ModelBinder.kt
littledice: You're right, left over from prev logic

littledice:
Thanks for these comments, I've tidied up the logic as suggested, including passing the value rather than the property around (not at all sure why I ever reversed that!) 

Re the question over the actual original bug, I'm in 2 minds about it. On the one hand, it would reduce code complexity to just stick the null check in. On the other hand, the current code future proofs usage of `RequiredWhen` against similar errors (as opposed to us remembering to null check non-nullable properties, which is an easy thing to overlook given our reliance on Kotlin's null safety!)

littledice:
Oh, re the user experience, I'm doing a `.distinct()` on the final error list

MartinEden:
There are a number of possible solutions here:

1. Only validate multi-property dependent validators (of which `RequiredWhen` is our only example currently) if single-property validators pass. This is straight-forward to implement, but as you say is a worse user experience, as the user may have to make multiple requests to get all feedback.
1. Make sure that the function used by `RequiredWhen` is safe to call, even when the object may be in an invalid state. This requires us to remember to write null-safe code is a context that looks like it doesn't need it. It has the advantage of simplicity.
1. Explicitly state which properties the function depends on to be non-null and verify this with reflection. This is the approach you have taken. It has the disadvantage of significant code complexity - particularly if we are actually going to fully implement it. At the moment, we have a half-way house between 1 and 3, as we only actually allow specification of one level of required properties, and then we insist that the whole of the subtree that picks out must have no nulls. This means that if we had another property of `BurdenEstimateSetType` that was unexpectedly null, but was not required for the `isStochastic` check, we would not return the `RequiredWhen` error, and we are back to the multi-stage validation problem. To truly implement this approach would require us to specify something like `RequiredWhen("isStochastic", "type.type")` and then recursively pop off each property and check it. This seems a bit overcomplicated! Worse, there's nothing to keep this dependent properties specification in sync with the `isStochastic` function, so we have exactly the same problem as 2: We have to remember to do the right null checks when we're writing it, we're just doing so via a reflection-based mechanism.
1. We could catch `NullReferenceException`s thrown by evaluating `RequiredWhen`'s function of interest, and return an empty list result if any are thrown. At first I shied away from this approach, as I feared that catching NREs indiscriminately would result in implementation bugs being mixed up with "not enough data to make a decision" cases. But actually, because Kotlin is a null-safe language, the *only* way a NRE could get thrown was in exactly the case we are trying to deal with: One where the user has submitted insufficient data to make a decision about `RequiredWhen`'s function of interest.

So how does 4 sound as an approach?

littledice:
Ok, you had me at this:  ` We still have to remember to do the right null checks when we're writing [the function that evaluates the `RequiredWhen` condition], we're just doing so via a reflection-based mechanism.`

4 is a good shout :+1:

littledice:
Ok, removed complicated null-checking code and a catching exceptions when calling the `RequireWhen` determiner. Unfortunately it's not enough to catch just `NullRefernceExceptions` - in this instance, when `type` is non null but `type.type` is, it throws an `IllegalArgumentException.` So I'm just catching any invocation that are thrown while executing that function.

littledice:
And finally, have made the qualified field names in the errors more comprehensible.

littledice:
:smile:

littledice:
Thanks for the patient reviewing!

PR #231, opened 2018-02-08
i1325: New endpoint

MartinEden:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-1325


littledice:
I think the order of these PRs is a bit backwards - should lock down the repo method first, and expose it via a new endpoint last. Especially as the validation will be the more complex bit of code, so might not get merged straight away. 

It's not toooo important, just nice to have master always in a deployable state (which I also violated this afternoon with my breaking db change :flushed:)

MartinEden:
Yeah good point. I did the new endpoint work before realizing that having that check was a good idea. So this is the order I did things in, rather than the best order for them to have been done in!

PR #233, opened 2018-02-08
i1325: Check burden estimate set is not empty before closing

MartinEden:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-1325


src/app/src/main/kotlin/org/vaccineimpact/api/app/errors/InvalidOperationError.kt
```diff
@@ -0,0 +1,7 @@
+package org.vaccineimpact.api.app.errors
+
+import org.vaccineimpact.api.models.ErrorInfo
+
+class InvalidOperationError(message: String) : MontaguError(400, listOf(
+        ErrorInfo("invalid-operation", message)
+))
```
MartinEden: Previously we had `OperationNotAllowedError`. This was only being used for impossible operations (e.g. state changes that can't happen), but somewhere along the way confusion has crept in: We were using status 401 Unauthorized, taking a different meaning of "not allowed".

So to clear things up I've renamed it, changed the error code to match, and changed the status to 400. There doesn't seem to be a more specific status code for "illegal operation".

The only user who might be affected is Rich. I'll liaise with him.

src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqBurdenEstimateRepository.kt
```diff
@@ -269,36 +272,35 @@ class JooqBurdenEstimateRepository(
...
     }
 
-    override fun closeBurdenEstimateSet(setId: Int)
+    override fun closeBurdenEstimateSet(setId: Int, groupId: String, touchstoneId: String, scenarioId: String)
     {
-        changeBurdenEstimateStatus(setId, BurdenEstimateSetStatus.COMPLETE)
+        // Check all the IDs match up
+        val (set, _) = getSetFromResponsibilityPath(groupId, touchstoneId, scenarioId, setId)
+        val helper = getEstimateWriter(set)
+        val isEmpty = dsl.fetchAny(helper.table, helper.setField.eq(setId)) == null
```
MartinEden: Here I'm using the `BurdenEstimateWriter` to look at the appropriate table, which means maybe we should rename it to `BurdenEstimateHelper`? Or maybe it's fine as is.

src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqBurdenEstimateRepository.kt
```diff
@@ -341,21 +346,24 @@ class JooqBurdenEstimateRepository(
...
         // changes fail, but the reverse is not true
         changeBurdenEstimateStatus(setId, BurdenEstimateSetStatus.EMPTY)
+        getEstimateWriter(set).clearEstimateSet(setId)
+    }
 
-        val type = set.type.type
-        if (type == BurdenEstimateSetTypeCode.STOCHASTIC)
+    private fun getEstimateWriter(set: BurdenEstimateSet): BurdenEstimateWriter
+    {
+        return if (set.type.type == BurdenEstimateSetTypeCode.STOCHASTIC)
```
littledice: could use the `.isStochastic()` helper

src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqBurdenEstimateRepository.kt
```diff
@@ -272,28 +275,30 @@ class JooqBurdenEstimateRepository(
...
         changeBurdenEstimateStatus(setId, BurdenEstimateSetStatus.PARTIAL)
         updateCurrentBurdenEstimateSet(responsibilityInfo.id, setId, type)
     }
 
     override fun closeBurdenEstimateSet(setId: Int, groupId: String, touchstoneId: String, scenarioId: String)
     {
         // Check all the IDs match up
-        getSetFromResponsibilityPath(groupId, touchstoneId, scenarioId, setId)
-        changeBurdenEstimateStatus(setId, BurdenEstimateSetStatus.COMPLETE)
+        val (set, _) = getSetFromResponsibilityPath(groupId, touchstoneId, scenarioId, setId)
```
littledice: Missed this on the last PR, but can we rename this function to `getSetWithResponsibilityInfo` so it's clearer what it returns?

src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqBurdenEstimateRepository.kt
```diff
@@ -272,28 +275,30 @@ class JooqBurdenEstimateRepository(
...
+        val helper = getEstimateWriter(set)
+        val isEmpty = dsl.fetchAny(helper.table, helper.setField.eq(setId)) == null
```
littledice: how about defining this as a function of `setId` on the estimate writer itself? I think fine to leave the naming as `...Writer` rather than `...Helper` though

src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqBurdenEstimateRepository.kt
MartinEden: Not quite, because it exists on `CreateBurdenEstimateSet`, not `BurdenEstimateSet`. However, I've now added a similar method: https://github.com/vimc/montagu-api/pull/233/commits/fe9971478f1a585caab62542489d2b38a0f927a7

You'll need to also merge this small webmodels PR https://github.com/vimc/montagu-webmodels/pull/41 and then return the API to webmodels master.

src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqBurdenEstimateRepository.kt
MartinEden: Done! https://github.com/vimc/montagu-api/pull/233/commits/44b8242fbdfc888deecea5a8eb101f0af929ff7e

src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqBurdenEstimateRepository.kt
MartinEden: Yeah, I wasn't sure if adding that was a good idea, because of polluting the purpose of `BurdenEstimateWriter` away from writing. But I think you're right. Done here: https://github.com/vimc/montagu-api/pull/233/commits/652e1ac19c7881b383a23e520d502acffed50470

src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqBurdenEstimateRepository.kt
littledice: Ah I meant `type.isStochastic` but extra helper method is even better :)

littledice:
These changes all look good, but there are still failing tests

PR #234, opened 2018-02-08
refresh db interface

littledice:
First merge https://github.com/vimc/montagu-db/pull/58


littledice:
In fact this can't be merged in isolation as it removes properties so won't compile

PR #237, opened 2018-02-09
i1423: Check Content-Type of file part is text/csv

MartinEden:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-1423


MartinEden:
This PR is the implementation, maintaining the status quo for architecture. Then the next PR refactors things to make them nicer.

PR #242, opened 2018-02-12
i1279: Verify new style onetime token

MartinEden:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-1279


src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/OneTimeLinkController.kt
```diff
@@ -124,7 +126,8 @@ class OneTimeLinkController(
 
         val claims = try
         {
-            tokenHelper.verify(token, TokenType.LEGACY_ONETIME)
+            tokenHelper.verify(token, TokenType.LEGACY_ONETIME,
+                    NoopOneTimeTokenChecker()) // token has already been checked in previous step
```
MartinEden: This is not that nice, but it's just to keep supporting the old style onetime tokens. It will go away once those are deprecated.

src/app/src/main/kotlin/org/vaccineimpact/api/app/security/JooqOneTimeTokenChecker.kt
```diff
@@ -0,0 +1,15 @@
...
+class JooqOneTimeTokenChecker(private val repositoryFactory: RepositoryFactory) : OneTimeTokenChecker
+{
+    override fun checkToken(token: String): Boolean
+    {
+        // This transaction is immediately committed, regardless of result
+        return repositoryFactory.inTransaction { repos ->
+            repos.token.validateOneTimeToken(token)
+        }
+    }
+}
```
MartinEden: By wrapping this logic in a simple interface (`OneTimeTokenChecker`) it allows the code to in the security module to remain blissfully unaware of Jooq.

src/app/src/main/kotlin/org/vaccineimpact/api/app/security/JWTParameterClient.kt
```diff
@@ -0,0 +1,28 @@
...
+import org.vaccineimpact.api.security.WebTokenHelper
+
+// This client receives the token as TokenCredentials and stores the result as JwtProfile
+class JWTParameterClient(helper: WebTokenHelper, oneTimeTokenChecker: OneTimeTokenChecker)
+    : ParameterClient("access_token", OneTimeTokenAuthenticator(helper, oneTimeTokenChecker))
+{
+    init
+    {
+        @Suppress("UsePropertyAccessSyntax")
+        this.setSupportGetRequest(true)
```
littledice: Why not use property access syntax?

src/app/src/main/kotlin/org/vaccineimpact/api/app/security/JooqOneTimeTokenChecker.kt
littledice: :+1:

src/security/src/main/kotlin/org/vaccineimpact/api/security/OneTimeTokenAuthenticator.kt
```diff
@@ -0,0 +1,36 @@
...
+        {
+            throw CredentialsException("Token has already been used (or never existed)")
+        }
+    }
+
+    override fun handleUrlAttribute(credentials: TokenCredentials, jwt: JWT)
+    {
+        val claims = jwt.jwtClaimsSet
+        val url = claims.getClaim("url")
+        @Suppress("UselessCallOnNotNull")
```
littledice: I see this is the same as the reporting API, but shall we just rewrite in a way that doesn't cause warnings? Looks like the `if` condition is equivalent to:
```
url == null || url.toString().isEmpty()
```

src/app/src/main/kotlin/org/vaccineimpact/api/app/security/JWTParameterClient.kt
MartinEden: Because then it reads `isSupportGetRequest = true`, which feels less readable to me here. The `is` prefix makes sense when it's something like `getEnabled`/`setEnabled` and it changes to `isEnabled`. But not so much here.

src/security/src/main/kotlin/org/vaccineimpact/api/security/OneTimeTokenAuthenticator.kt
MartinEden: How's this version (https://github.com/vimc/montagu-api/pull/242/commits/5ba6f201185e25831558bb962f54f48eb1be964a); I think it's clearer. The important thing is that we handle the case (without throwing a weird exception) where the URL claim is missing, or a date, or something strange like that.

src/app/src/main/kotlin/org/vaccineimpact/api/app/security/JWTParameterClient.kt
littledice: I thought this was why. That's fine

PR #251, opened 2018-03-02
i1108: Update to not use deprecated methods

MartinEden:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-1108


src/databaseInterface/src/main/java/org/vaccineimpact/api/db/Routines.java
```diff
@@ -69,27 +68,6 @@ public static void enableTrigger(Configuration configuration, String tableName,
         p.execute(configuration);
     }
 
-    /**
```
littledice: Isn't this file auto-generated? Similarly the one below?

littledice:
Also TC is failing :(

src/databaseInterface/src/main/java/org/vaccineimpact/api/db/Routines.java
MartinEden: Yup. I've changed the xml file that configures how it is generated. See main-database.xml.

MartinEden:
Fixed in TeamCity

src/build.gradle
```diff
@@ -97,7 +97,12 @@ subprojects {
...
     }
 
-    tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).all {
+    compileKotlin {
+        kotlinOptions {
+            jvmTarget = "1.8"
+        }
+    }
+    compileTestKotlin {
         kotlinOptions {
```
MartinEden: This change isn't strictly necessary - it's just a refactor to improve readability.

PR #253, opened 2018-03-02
i1568: Track latest openJDK image

MartinEden:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-1568

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,10 @@ RUN add-apt-repository \
    $(lsb_release -cs) \
    stable"
 RUN apt-get update
-RUN apt-get install -y docker-ce=17.03.0~ce-0~debian-jessie
+# This outputs available Docker versions - useful for choosing a new version
+# if the pinned version is deprecated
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -13,7 +13,10 @@ RUN add-apt-repository \
    $(lsb_release -cs) \
    stable"
 RUN apt-get update
-RUN apt-get install -y docker-ce=17.03.0~ce-0~debian-jessie
+# This outputs available Docker versions - useful for choosing a new version
+# if the pinned version is deprecated
+RUN apt-cache madison docker-ce
+RUN apt-get install -y docker-ce=17.12.1~ce-0~debian
```
MartinEden: So the official OpenJDK image has changed from Ubuntu to vanilla Debian. But so long as all our tests pass, this feels like it should be a safe change to make. As far as we're concerned we're just running a Java program in a Java docker container, and the OS of the docker image shouldn't bother us.

PR #260, opened 2018-04-17
I1671 - endpoint to save user confidentiality agreement

littledice:
And one to GET whether the logged in user has signed it.

First merge https://github.com/vimc/montagu-db/pull/70 and return db to master.


littledice:
Sure, done. On that note there is no unique constraint in the database table and the endpoint at the moment just adds a new entry each time, with a new timestamp. In practice the webapps widget doesn't let you un-check/re-check the box once it's ticked, so users will only make one call to this endpoint.

src/databaseInterface/src/main/java/org/vaccineimpact/api/db/Sequences.java
```diff
@@ -26,22 +26,22 @@
     /**
      * The sequence <code>public.api_access_log_id_seq</code>
      */
-    public static final Sequence<Long> API_ACCESS_LOG_ID_SEQ = new SequenceImpl<Long>("api_access_log_id_seq", Public.PUBLIC, org.jooq.impl.SQLDataType.BIGINT.nullable(false));
```
littledice: I just re-ran the db interface generator, using db master (as there were conflicts) and all of these changes appeared - any idea why? I would have expected these to be introduced in `i1480`?

PR #261, opened 2018-04-18
parse floats 

littledice:
and change exception catching logic so throws more informative error

### DIFF
--- a/src/blackboxTests/src/test/kotlin/org/vaccineimpact/api/blackboxTests/helpers/PermissionChecker.kt
+++ b/src/blackboxTests/src/test/kotlin/org/vaccineimpact/api/blackboxTests/helpers/PermissionChecker.kt
@@ -73,7 +73,7 @@ class PermissionChecker(
             val scope = permission.scope as Scope.Specific
 
             println("Checking that same permission with different scope will not satisfy the requirement")
-            val badPermission = ReifiedPermission(permission.name, Scope.Specific(scope.scopePrefix, "bad-id"))
+            val badPermission = ReifiedPermission(permission.name, Scope.Specific(scope.databaseScopePrefix, "bad-id"))
             checkThesePermissionsAreInsufficient(limitedPermissions + badPermission,
                     expectedProblem, assertionText)
 
--- a/src/blackboxTests/src/test/kotlin/org/vaccineimpact/api/blackboxTests/tests/CoverageTests.kt
+++ b/src/blackboxTests/src/test/kotlin/org/vaccineimpact/api/blackboxTests/tests/CoverageTests.kt
@@ -120,22 +120,22 @@ class CoverageTests : DatabaseTest()
 
         val expected = listOf(
                 TestWideCoverageRow(scenarioId, "First", "AF", "no gavi", "routine",
-                        "AAA", "AAA-Name", 2.toDecimal(), 4.toDecimal(), "<NA>", yearMap),
+                        "AAA", "AAA-Name", 2, 4, "<NA>", yearMap),
                 // first order by vaccine
                 TestWideCoverageRow(scenarioId, "Second", "BF", "no gavi", "campaign",
-                        "AAA", "AAA-Name", 1.toDecimal(), 2.toDecimal(), "<NA>", yearMap),
+                        "AAA", "AAA-Name", 1, 2, "<NA>", yearMap),
                 // then by activity type
                 TestWideCoverageRow(scenarioId, "Third", "BF", "no gavi", "routine",
-                        "AAA", "AAA-Name", 1.toDecimal(), 2.toDecimal(), "<NA>", yearMap),
+                        "AAA", "AAA-Name", 1, 2, "<NA>", yearMap),
                 // then by country
                 TestWideCoverageRow(scenarioId, "Third", "BF", "no gavi", "routine",
-                        "BBB", "BBB-Name", 1.toDecimal(), 2.toDecimal(), "<NA>", yearMap),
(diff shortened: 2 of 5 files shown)


src/blackboxTests/src/test/kotlin/org/vaccineimpact/api/blackboxTests/helpers/PermissionChecker.kt
```diff
@@ -73,7 +73,7 @@ class PermissionChecker(
             val scope = permission.scope as Scope.Specific
 
             println("Checking that same permission with different scope will not satisfy the requirement")
-            val badPermission = ReifiedPermission(permission.name, Scope.Specific(scope.scopePrefix, "bad-id"))
+            val badPermission = ReifiedPermission(permission.name, Scope.Specific(scope.databaseScopePrefix, "bad-id"))
```
littledice: Not sure why/how this change made it in, but there is no longer a `val` called `scopePrefix` in `Scope.Specific`

src/blackboxTests/src/test/kotlin/org/vaccineimpact/api/blackboxTests/tests/CoverageTests.kt
```diff
@@ -213,8 +213,8 @@ class CoverageTests : DatabaseTest()
             val activityType: String,
             val countryCode: String,
             val country: String,
-            val ageFirst: BigDecimal?,
-            val ageLast: BigDecimal?,
+            val ageFirst: Int?,
+            val ageLast: Int?,
```
littledice: This is a test class that is only used here

src/serialization/src/main/kotlin/org/vaccineimpact/api/serialization/Deserializer.kt
```diff
@@ -32,7 +31,7 @@ class Deserializer
         {
             String::class.createType() -> raw
             Int::class.createType() -> raw.toInt()
-            BigDecimal::class.createType() -> BigDecimal(raw)
+            Float::class.createType() -> raw.toFloat()
```
MartinEden: Don't we want to be able to handle both of these types? So just add the float code, but no need to remove the decimal.

src/serialization/src/main/kotlin/org/vaccineimpact/api/serialization/DataTableDeserializer.kt
```diff
@@ -85,11 +85,18 @@ open class DataTableDeserializer<out T>(
         }
         catch (e: Exception)
         {
+            if (e is UnsupportedOperationException)
+            {
+                throw e
+            }
```
MartinEden: Could you write this as a separate catch block that comes before the general case catch, rather than as an if statement?

src/blackboxTests/src/test/kotlin/org/vaccineimpact/api/blackboxTests/tests/CoverageTests.kt
```diff
@@ -120,22 +120,22 @@ class CoverageTests : DatabaseTest()
 
         val expected = listOf(
                 TestWideCoverageRow(scenarioId, "First", "AF", "no gavi", "routine",
-                        "AAA", "AAA-Name", 2.toDecimal(), 4.toDecimal(), "<NA>", yearMap),
+                        "AAA", "AAA-Name", 2, 4, "<NA>", yearMap),
```
MartinEden: If you don't remove the ability to handle decimals (see below) then I think these changes become unnecessary.

src/blackboxTests/src/test/kotlin/org/vaccineimpact/api/blackboxTests/helpers/PermissionChecker.kt
MartinEden: Weird

src/serialization/src/main/kotlin/org/vaccineimpact/api/serialization/Deserializer.kt
littledice: So I wondered the same about removing the ability to parse decimals, but I think we should for now, given none of our target types at the moment have decimals in them (and maybe now none of our database data structures?) - if we need to parse decimals again down the line, we can add it back in.

src/serialization/src/main/kotlin/org/vaccineimpact/api/serialization/DataTableDeserializer.kt
littledice: It can't come before this catch block, otherwise it'll still be caught by this one!

src/blackboxTests/src/test/kotlin/org/vaccineimpact/api/blackboxTests/tests/CoverageTests.kt
littledice: This is just a test class, no reason for it to use decimals.

src/serialization/src/main/kotlin/org/vaccineimpact/api/serialization/Deserializer.kt
MartinEden: We have decimals in the db - coverage and demographics - but it's true that the API doesn't write to those at the moment.

src/serialization/src/main/kotlin/org/vaccineimpact/api/serialization/DataTableDeserializer.kt
MartinEden: Good point :)

PR #264, opened 2018-04-23
I1695 - libsodium

littledice:
Move libsodium docker logic out of the API and instead pull the libsodium image from the registry where needed.


src/app/build.gradle
```diff
@@ -51,9 +57,9 @@ dependencies {
 }
 
 distDocker {
-    baseImage = "libsodium"
+    baseImage = base_image
     dependsOn = [build]
-    push = true
+    push = false
```
MartinEden: Why has this changed?

blackbox.Dockerfile
```diff
@@ -1,4 +1,5 @@
-FROM libsodium
+ARG libsodium_image=docker.montagu.dide.ic.ac.uk:5000/libsodium:i1695
```
MartinEden: Must remember to return this to master

src/app/build.gradle
littledice: oops, local testing

littledice:
Like read it in from a text file? I doubt it will change very often (or ever?) so I'd propose keeping it simple as is.

MartinEden:
Okay

PR #265, opened 2018-06-08
Upgrade Gradle to work on Martin's machine

MartinEden:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-1827


src/databaseInterface/build.gradle
```diff
@@ -4,4 +4,5 @@ dependencies {
     compile "org.jooq:jooq:3.10.5"
     compile "org.jooq:jooq-meta:3.10.5"
     compile "org.apache.commons:commons-lang3:3.7"
+    compile "javax.annotation:javax.annotation-api:1.3.2"
```
MartinEden: This seems to now need to be included manually, whereas previously it was bundled as part of the Java 8 stdlib, I think?

PR #268, opened 2018-06-11
i1825: Change endpoints

MartinEden:
First merge #267 

https://vimc.myjetbrains.com/youtrack/issue/VIMC-1825

This changes the two endpoints to return touchstones as well as touchstone versions. This should not be merged until the portals have been updated.


src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/TouchstoneController.kt
```diff
@@ -23,30 +23,28 @@ class TouchstoneController(
...
-        var touchstones = repo.touchstones.all()
-        touchstones = touchstones.filter { context.isAllowedToSeeTouchstone(it.status) }
-        return touchstones.toList()
+        return repo.getTouchstones().filterByPermission(context)
     }
 
     fun getScenarios(): List<ScenarioAndCoverageSets>
     {
-        val touchstone = touchstone(context, repo)
+        val touchstone = touchstoneVersion(context, repo)
```
littledice: sorry to be super pedantic, but a few more variables to rename here

src/app/src/test/kotlin/org/vaccineimpact/api/tests/controllers/ResponsibilityControllerTests.kt
```diff
@@ -110,8 +110,10 @@ class ResponsibilityControllerTests : MontaguTests()
 
 
     private val mockTouchstones = listOf(
-            TouchstoneVersion("touchstone-1", "touchstone", 1, "Description", TouchstoneStatus.OPEN),
-            TouchstoneVersion("touchstone-bad", "touchstone", 1, "not open", TouchstoneStatus.IN_PREPARATION)
+            Touchstone("touchstone", "description", "comment", listOf(
```
littledice: We also need to update the test that we're only returning open touchstone versions, `does not return in preparation touchstones if user has no permission to read prepared touchstones`

PR #271, opened 2018-06-12
I1838 - Endpoint to get responsibilities in a touchstone

littledice:
Involved some refactoring - pulled out methods from the `ModellingGroupRepository` into a `ResponsibilitiesRepository` so they could be re-used. Reminded me that we should really think about adding a logic layer between the controllers and the repos!

1 webmodels PR along with this one: https://github.com/vimc/montagu-webmodels/pull/49

### DIFF
--- a/src/app/src/main/kotlin/org/vaccineimpact/api/app/app_start/route_config/TouchstoneRouteConfig.kt
+++ b/src/app/src/main/kotlin/org/vaccineimpact/api/app/app_start/route_config/TouchstoneRouteConfig.kt
@@ -13,11 +13,16 @@ object TouchstoneRouteConfig : RouteConfig
     private val permissions = setOf("*/touchstones.read")
     private val scenarioPermissions = permissions + setOf("*/scenarios.read", "*/coverage.read")
     private val demographicPermissions = permissions + setOf("*/demographics.read")
+    private val responsibilityPermissions = scenarioPermissions + setOf("*/responsibilities.read")
 
     override val endpoints: List<EndpointDefinition> = listOf(
             Endpoint(baseUrl, controller, "getTouchstones")
                     .json()
                     .secure(permissions),
+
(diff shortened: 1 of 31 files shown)


src/app/src/main/kotlin/org/vaccineimpact/api/app/app_start/route_config/TouchstoneRouteConfig.kt
```diff
@@ -13,12 +13,17 @@ object TouchstoneRouteConfig : RouteConfig
     private val permissions = setOf("*/touchstones.read")
     private val scenarioPermissions = permissions + setOf("*/scenarios.read", "*/coverage.read")
     private val demographicPermissions = permissions + setOf("*/demographics.read")
+    private val responsibilityPermissions = permissions + setOf("*/scenarios.read", "*/responsibilities.read")
```
MartinEden: I don't think this matches the spec. Unless I'm reading the wrong bit, above you said "Required permissions: `touchstones.read`, `responsibilities.read`"

src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/ResponsibilityController.kt
```diff
@@ -14,17 +15,18 @@ import org.vaccineimpact.api.models.TouchstoneVersion
 
 class ResponsibilityController(
```
MartinEden: Maybe rename this to `GroupResponsibilityController` now that there are responsibilities on another controller.

src/app/src/main/kotlin/org/vaccineimpact/api/app/app_start/route_config/TouchstoneRouteConfig.kt
littledice: Ah you're right. I think this stricter condition is correct, because the scenarios are returned with the responsibilities. Will update spec.

MartinEden:
Looks fine now; just the merge conflicts.

PR #273, opened 2018-06-15
I1840 - create new modelling group

littledice:


docs/schemas/ModellingGroupCreation.schema.json
```diff
@@ -0,0 +1,12 @@
+{
```
littledice: The alternative of course to this new schema (and new data class of the same name) is that we just add the properties "pi" and "institution" to the `ModellingGroup` data class and schema. :woman_shrugging:

MartinEden: No, I think it's really good to have separate objects here.

PR #274, opened 2018-06-15
i1866: GZip bearer token

MartinEden:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-1866

This must be merged iff https://github.com/vimc/montagu-webapps/pull/256 is also merged at the same time.


src/security/src/main/kotlin/org/vaccineimpact/api/security/MontaguTokenAuthenticator.kt
```diff
@@ -3,13 +3,19 @@ package org.vaccineimpact.api.security
...
 import org.pac4j.core.exception.CredentialsException
+import org.pac4j.core.profile.CommonProfile
 import org.pac4j.jwt.credentials.authenticator.JwtAuthenticator
 
 open class MontaguTokenAuthenticator(
         private val tokenHelper: WebTokenHelper,
         private val expectedType: TokenType
 ) : JwtAuthenticator(tokenHelper.signatureConfiguration)
 {
+    override fun validateToken(token: String?): CommonProfile?
```
littledice: It'd be good to add a test for this, `canValidateDeflatedToken`?

src/security/src/main/kotlin/org/vaccineimpact/api/security/WebTokenHelper.kt
```diff
@@ -21,7 +21,7 @@ open class WebTokenHelper(keyPair: KeyPair,
     open val defaultLifespan: Duration = Duration.ofSeconds(Config["token.lifespan"].toLong())
     val issuer = Config["token.issuer"]
     val signatureConfiguration = RSASignatureConfiguration(keyPair)
-    val generator = JwtGenerator<CommonProfile>(signatureConfiguration)
+    val generator: TokenGenerator = CompressedJwtGenerator<CommonProfile>(signatureConfiguration)
```
littledice: I'm not 100% sure the compression of the token should happen inside here, as it's not obvious that by calling `webTokenHelper.generateToken` you should be expecting a gzipped token back. 

Intuitively it'd make more sense to me to deflate the token right before returning it to the user (i.e. in the controller), and then you also wouldn't have to call `inflate` in the repository before saving the token.

Either that or seeing as the `CompressedJWTGenerator` is hard-coded into this class (which also feels wrong to me, you know how militant I am about dependency injection :P)  we could just rename the methods here to `generateDeflatedToken` etc to make it clear what type of thing we're expecting back.

src/security/src/main/kotlin/org/vaccineimpact/api/security/MontaguTokenAuthenticator.kt
MartinEden: Currently this class is implicitly tested by `WebTokenHelperTests`. I think this is a reasonable approach to keep taking, in which case `can generate bearer token` tests this path.

src/security/src/main/kotlin/org/vaccineimpact/api/security/WebTokenHelper.kt
MartinEden: I've opened https://github.com/vimc/montagu-api/pull/276, which contains a different approach. As you suggest, it moves the compression outside of `WebTokenHelper`. This means compression happens in more places, and is a bigger change, but is perhaps clearer. I've also introduced a bit of type marking to help us keep track of what's going on, although this is inherently limited by the fact that when we interface with the outside world everything has to ultimately go in and out of strings.

Take a look at that PR and see if you prefer it.

littledice: Ok, bear with me, I'm really not sure which approach I prefer!

PR #281, opened 2018-07-17
I1951 - generate csv template

littledice:
First merge https://github.com/vimc/montagu-api/pull/283

Associated PR https://github.com/vimc/montagu-webmodels/pull/52/

This does actually also add the endpoint just because that was the easiest way to test it! But there's another PR that finishes off the work around the endpoint, i.e. adds it to the docs and makes it accept a query parameter to specify the type of template (stochastic/central)


src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqExpectationsRepository.kt
```diff
@@ -4,27 +4,30 @@ import org.jooq.DSLContext
...
 import org.vaccineimpact.api.db.tables.records.BurdenEstimateExpectationRecord
-import org.vaccineimpact.api.models.CohortRestriction
-import org.vaccineimpact.api.models.Country
-import org.vaccineimpact.api.models.Expectations
-import org.vaccineimpact.api.models.YearRange
+import org.vaccineimpact.api.models.*
 
-class JooqExpectationsRepository(dsl: DSLContext) : JooqRepository(dsl), ExpectationsRepository
+class JooqExpectationsRepository(dsl: DSLContext,
+                                 private val responsibilitiesRepository: ResponsibilitiesRepository)
```
littledice: Every time I touch the repos I feel keen to refactor; maybe we could bring in the ticket to introduce a logic layer into this sprint?

MartinEden:
Also, I'm not sure that the logic for generating the burden estimate sequence should live in the `Expectations` model - maybe in a separate `BurdenEstimateTemplateGenerator` class that takes an `Expectations` object as an input?

src/app/src/main/kotlin/org/vaccineimpact/api/app/app_start/route_config/ResponsibilityRouteConfig.kt
```diff
@@ -29,6 +26,11 @@ object ResponsibilityRouteConfig : RouteConfig
 
             Endpoint("$baseUrl/:touchstone-version-id/:scenario-id/", controller, "getResponsibility")
                     .json()
+                    .secure(permissions),
+
+            Endpoint("$baseUrl/:touchstone-version-id/:scenario-id/template/", controller, "getTemplate")
```
littledice: This will change in future PR, but here for testing purposes

littledice:
Ok, tracking actual expected rows as suggested, with a new class for generating an empty csv. I do think that the expected rows logic makes sense conceptually in the `Expectations` class though, as it's just another way of describing the expectations.

src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/GroupResponsibilityController.kt
```diff
@@ -56,12 +59,17 @@ class GroupResponsibilityController(
...
 
-    fun getTemplate()
+    fun getTemplate(): StreamSerializable<Any?>
     {
         val path = ResponsibilityPath(context)
         modellingGroupRepo.getModellingGroup(path.groupId)
         val expectations = expectationsLogic.getExpectationsForResponsibility(path.groupId,
                 path.touchstoneVersionId, path.scenarioId)
+
+        val headers = arrayOf("disease", "year", "age", "country", "countryName", "cohortSize") +
```
littledice: Wasn't sure whether to pull these from `BurdenEstimate` by reflection, or just hard-code here :woman_shrugging:

MartinEden: How about you make `EmptyDataTable` a bit more like `DataTable` - give it a type to derive headers from, and reuse the logic for that? You can still then just generate a number of blank rows off the back of that.

littledice:
Ok, now extending the `FlexibleDataTable` class and using the `BurdenEstimate` models for headers. I think this is nicer :)

PR #282, opened 2018-07-17
I1950 - deliver central or stochastic template, depending on query param

littledice:
Associated PR: https://github.com/vimc/montagu-webmodels/pull/51

But first merge https://github.com/vimc/montagu-api/pull/281


docs/spec/Responsibilities.md
```diff
@@ -144,6 +144,17 @@ Schema: [`ResponsibilityAndTouchstone.schema.json`](../schemas/ResponsibilityAnd
         }
     }
 
+## GET /modelling-groups/{modelling-group-id}/responsibilities/{touchstone-id}/{scenario-id}/template/
```
MartinEden: This doesn't match up with what's implemented.

src/app/src/main/kotlin/org/vaccineimpact/api/app/logic/ExpectationsLogic.kt
```diff
@@ -2,28 +2,26 @@ package org.vaccineimpact.api.app.logic
...
 {
-    override fun getExpectationsForResponsibility(groupId: String, touchstoneVersionId: String, scenarioId: String): Expectations
+    override fun getExpectations(groupId: String, touchstoneVersionId: String, expectationId: Int):
+            Expectations
     {
         modellingGroupRepository.getModellingGroup(groupId) // throws if group does not exist
         touchstoneRepository.touchstoneVersions.get(touchstoneVersionId) // throws if touchstone version does not exist
-        val responsibilityId = responsibilitiesRepository.getResponsibilityId(groupId, touchstoneVersionId, scenarioId)
-        return expectationsRepository.getExpectationsForResponsibility(responsibilityId)
+        return expectationsRepository.getExpectations(expectationId)
```
MartinEden: I'm not sure we are currently checking that this expectation belongs to this modelling group and touchstone.

littledice: Just realised you are doing some refactoring to return expectations with responsibilities, so I'll wait for that to go in as will presumably make this check easier.

littledice:
I'd be grateful if you could look over the `ExpectationsLogicTests` class and just check that the tests are correct. The two methods `getExpectationsForResponsibility` and `getResponsibilityWithExpectations` have the same signature and very similar names, which makes that test class very confusing.

MartinEden:
I'll look at this properly tomorrow morning, but: Perhaps we should get rid of `getExpectationsForResponsibility` and just always use `getResponsibilityWithExpectations`? It returns a super-set of the same data, which I guess is slightly inefficient, but not enough to cause us problems I think. Then there is less to reason about and test.

littledice:
Oh, I just realised that `getExpectationsForResponsibility` is totally obsolete. I wrote it specifically for this endpoint, which is now using the new method :woman_facepalming:

PR #283, opened 2018-07-17
change signature of getExpectations method

littledice:


src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/ExpectationsRepository.kt
```diff
@@ -4,5 +4,7 @@ import org.vaccineimpact.api.models.Expectations
 
 interface ExpectationsRepository : Repository
 {
-    fun getExpectationsForResponsibility(responsibilityId: Int): Expectations
+    fun getExpectationsForResponsibility(groupId: String,
```
littledice: I changed the signature of this method since this is how the ids come in to the endpoint, and is consistent with how we do things elsewhere.

src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqResponsibilitiesRepository.kt
```diff
@@ -123,6 +123,25 @@ class JooqResponsibilitiesRepository(
         }
     }
 
+    override fun getResponsibilityId(groupId: String, touchstoneVersionId: String, scenarioId: String): Int
+    {
+        val responsibilitySet = getResponsibilitySet(groupId, touchstoneVersionId)
```
MartinEden: Should throw errors if group or touchstone don't exist?

src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqResponsibilitiesRepository.kt
```diff
@@ -123,6 +123,25 @@ class JooqResponsibilitiesRepository(
...
+        if (responsibilitySet != null)
+        {
+            val id = dsl.select(Tables.RESPONSIBILITY.ID)
+                    .fromJoinPath(Tables.RESPONSIBILITY, Tables.SCENARIO, Tables.SCENARIO_DESCRIPTION)
+                    .where(Tables.RESPONSIBILITY.RESPONSIBILITY_SET.eq(responsibilitySet[Tables.RESPONSIBILITY_SET.ID])
+                    .and(Tables.SCENARIO_DESCRIPTION.ID.eq(scenarioId))
+            ).singleOrNull() ?: throw UnknownObjectError(scenarioId, "responsibility")
```
MartinEden: Maybe use `fetchAnyInto` to do the `singleOrNull` and `into` all in one step? Also, the indentation is a bit screwy here.

src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqResponsibilitiesRepository.kt
littledice: Actually that logic should now live in the logic class! But yes you're right, I'll add it there :+1:

src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqResponsibilitiesRepository.kt
littledice: Does `fetchAnyInto` return null if there is no record or does it throw? Wasn't sure

MartinEden: Not sure, tbh

PR #287, opened 2018-07-19
i1965: Return more data from the get responsibility endpoint

MartinEden:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-1965

Need to return models to master after merging ttps://github.com/vimc/montagu-webmodels/pull/54


src/databaseTests/src/test/kotlin/org/vaccineimpact/api/databaseTests/tests/responsibilitiesRepository/GetResponsibilityTests.kt
```diff
@@ -149,23 +149,24 @@ class GetResponsibilityTests : ResponsibilitiesRepositoryTests()
...
     fun `can get responsibility`()
     {
         given {
             it.addGroup("group-1", "description")
             it.addTouchstoneVersion("touchstone", 1, "description", "open", addTouchstone = true)
             it.addScenarioDescription("scenario-1", "description", "disease", addDisease = true)
             val setId = it.addResponsibilitySet("group-1", "touchstone-1", "incomplete")
-            it.addResponsibility(setId, "touchstone-1", "scenario-1")
+            val responsibilityId = it.addResponsibility(setId, "touchstone-1", "scenario-1")
+            it.addExpectations(responsibilityId)
```
littledice: Is this neccessary? This method should work without expectations right?

MartinEden: Correct. I changed that test and then changed it back, but clearly left some bits and pieces lying around. I'll fix it

MartinEden:
I've made some changes. I haven't updated the tests yet, I wanted to first get your feedback on the logic refactor. Note that there's another small PR for webmodels that this now depends on: https://github.com/vimc/montagu-webmodels/pull/55

littledice:
Cool, yes this all seems good :+1: 
Although one thing it has me pondering now is whether we should have a concept of DTO models as separate from internal ones! I know there was a conscious choice earlier in the project not to have that separation, but maybe worth revisiting at some point to make sure it still makes sense.

MartinEden:
Yes, I was thinking the same. Once we're having two layers, we should have two layers of models, or at least a separate namespace to put models that are purely used in the repositories layer, and are never exposed to the public.

PR #291, opened 2018-07-20
i1969: Add expectations to two existing endpoints

MartinEden:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-1969

### DIFF
--- a/src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/GroupResponsibilityController.kt
+++ b/src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/GroupResponsibilityController.kt
@@ -14,7 +14,7 @@ import org.vaccineimpact.api.app.security.filterByPermission
 import org.vaccineimpact.api.models.responsibilities.ResponsibilityDetails
 import org.vaccineimpact.api.models.BurdenEstimate
 import org.vaccineimpact.api.models.Touchstone
-import org.vaccineimpact.api.models.responsibilities.ResponsibilitiesWithExpectations
+import org.vaccineimpact.api.models.responsibilities.ResponsibilitySetWithExpectations
 import org.vaccineimpact.api.serialization.EmptyDataTable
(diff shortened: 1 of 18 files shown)


src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/GroupResponsibilityController.kt
```diff
@@ -37,17 +37,14 @@ class GroupResponsibilityController(
...
         val groupId = groupId(context)
         val touchstoneVersionId = context.params(":touchstone-version-id")
         val filterParameters = ScenarioFilterParameters.fromContext(context)
-
-        modellingGroupRepo.getModellingGroup(groupId)
-        val data = responsibilitiesRepo
-                .getResponsibilitiesForGroupAndTouchstone(groupId, touchstoneVersionId, filterParameters)
-        context.checkIsAllowedToSeeTouchstone(touchstoneVersionId, data.touchstoneStatus)
-        return data.responsibilitySet
+        val (responsibilities, touchstoneStatus) = expectationsLogic.getResponsibilitySetWithExpectations(groupId, touchstoneVersionId, filterParameters)
```
littledice: I know we were doing this before, but it seems kind of hack-y to be calling a method called `getResponsibilitySetWithExpectations` and actually getting back the touchstone status as a sort of bolt on. 

Can we just make a separate request to get the touchstone status and have this new logic method do what it says on the tin? From what I can see the underlying `ResponsibilitiesRepository` method just calls `touchstoneRepository.touchstoneVersions.get(touchstoneVersionId)` to get the status and pass it back. We may as well just call that directly here.

MartinEden: Yes you're right. I tried to figure out how to simplify this before, but got confused. This is much better. Note that there is now an extra webmodels PR: https://github.com/vimc/montagu-webmodels/pull/57

PR #294, opened 2018-07-25
return semi-populated rows for templates

littledice:
So, the only thing stopping us using the `FlexibleDataTable` class is the fact that nulls get serialized to `<NA>` which we don't want in the case of templates.

The serializer was a method level argument of the `serialize` method on the `StreamSerializable` interface. This meant we could have a single point where we chose a serializer - when we invoke that method for an endpoint of type `StreamSerializable`, which ensures consistent serialization of these endpoints. However I can't think of a better way forward other than taking a serializer as a class level argument when constructing a `DataTable` rather than a method level argument.

See what you think of this refactor.

Expectations logic here: https://github.com/vimc/montagu-webmodels/pull/59


src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/GroupResponsibilityController.kt
```diff
@@ -68,14 +69,15 @@ class GroupResponsibilityController(
...
         {
-            EmptyDataTable.new<BurdenEstimate>(rowCount, expectations.outcomes)
+            FlexibleDataTable.new(expectations.expectedCentralRows(expectationMapping.disease),
+                    expectations.outcomes, NullToEmptyStringSerializer.instance)
         }
         else
         {
-            EmptyDataTable.new<StochasticBurdenEstimate>(rowCount, expectations.outcomes)
+            FlexibleDataTable.new(expectations.expectedStochasticRows(expectationMapping.disease),
+                    expectations.outcomes, NullToEmptyStringSerializer.instance)
```
MartinEden: Again, seems like more of this code could be shared between the two paths? In fact, could `type` just be an argument to a single method on the `Expectations` object?

src/app/src/test/kotlin/org/vaccineimpact/api/tests/extensions/ExpectationsTests.kt
```diff
@@ -21,12 +21,81 @@ class ExpectationsTests : MontaguTests()
...
+                CohortRestriction(null, null),
+                longCountryList,
+                expectedOutcomes
+        )
+
+        val result = expectations.expectedCentralRows("YF").toList()
+
+        assertThat(result.all { it.cohortSize == null }).isTrue()
+        assertThat(result.all { it.outcomes.all { it.value == null } }).isTrue()
+    }
```
MartinEden: It would also be really good to have some tests that generate a much smaller number of rows (e.g. 2 years, 2 countries, 2 ages = 8 rows) and checks that you can get back exactly the objects expected, given that contents matters more now, rather than just count.

src/serialization/src/main/kotlin/org/vaccineimpact/api/serialization/MontaguSerializer.kt
```diff
@@ -23,6 +23,21 @@ interface Serializer
     val gson: Gson
 }
 
+class NullToEmptyStringSerializer(serializer: Serializer = MontaguSerializer.instance): Serializer by serializer {
+
+    override fun serializeValueForCSV(value: Any?) = when (value)
+    {
+        null -> ""
```
MartinEden: Could refactor the base class so you only have to override the `null` branch of the when, rather than the whole thing. e.g. by introducing a `serializeNull` method on the base class and then just overriding that in this class?

src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/GroupResponsibilityController.kt
littledice: It could be, yes. See comment in `webmodels`. If you feel strongly I'll change, I just thought it was clearer this way - I think there's something a bit off about returning different types of things depending on a method argument!

PR #302, opened 2018-08-10
i2056: Allow result redirection in burden estimate upload

MartinEden:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-2056

One extra change here: We weren't actually using the one time link for *creating* burden estimate sets (as opposed to populating them) so rather than have to refactor that here, I just removed it. This means there's also a [tiny webmodels PR](https://github.com/vimc/montagu-webmodels/pull/61).


src/app/src/main/kotlin/org/vaccineimpact/api/app/ResultRedirector.kt
```diff
@@ -0,0 +1,60 @@
+package org.vaccineimpact.api.app
```
MartinEden: This code largely duplicates existing logic in `OneTimeLinkController`, but since that will be removed soon I didn't try and make them share code.

PR #303, opened 2018-08-15
add  new role to docs

littledice:


MartinEden:
Oops. Shouldn't have merged this - there's a bug now

PR #306, opened 2018-08-15
i2073: Issue cookie on login

MartinEden:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-2073


src/app/src/test/kotlin/org/vaccineimpact/api/tests/logic/UserLogicTests.kt
```diff
@@ -0,0 +1,35 @@
...
+    fun `getUserByUsername passes through to repository`()
+    {
+        val repo = mock<UserRepository>()
+        val sut = RepositoriesUserLogic(repo, mock())
+        sut.getUserByUsername("user")
+        verify(repo).getUserByUsername("user")
+    }
+
+    @Test
+    fun `logInAndGetToken gets token and updates last logged in`()
```
littledice: Looks like we had a test before that checked that `lastLoggedIn` wasn't updated on a failed login, but that's no longer under test? Unless I've missed it

MartinEden: This is still tested, just less directly: In the controller test, we check that the logic method `logInAndGetToken` is not called. And in the logic test we check that calling `logInAndGetToken` sets `lastLoggedIn`.

littledice: Ah yes, ok

MartinEden:
Make sure we return webmodels to master after merging https://github.com/vimc/montagu-webmodels/pull/62

PR #313, opened 2018-08-16
I2083 - remove all old onetime link code

littledice:
Just deletions and a bit of renaming.
Also: https://github.com/vimc/montagu-webmodels/pull/64


src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/PasswordController.kt
```diff
@@ -37,6 +36,8 @@ class PasswordController(
     fun setPasswordForUser(username: String): String
     {
         val password = context.postData<SetPassword>().password
+        logger.debug(password)
+        logger.debug(username)
```
MartinEden: Debug statements need removing - particularly since it has a password in it!

littledice:
Any remaining Blackbox tests must be for new style tokens, any tests pointing to the old endpoint would fail since all that code has been removed.

src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/PasswordController.kt
littledice: Oops, good spot!

MartinEden:
But in that case why not migrate the tests over to use new style tokens, rather than removing them?

littledice:
We used to have one test per download/upload endpoint, because each one required it's own endpoint and route config. The new onetime access tokens are a global auth mechanism that every endpoint supports, so it's sufficient to test it once (and there is just one endpoint to retrieve a token, as opposed to one per download/upload endpoint.)

littledice:
If you think any of the deleted tests are testing functionality that isn't already covered, could you point them out?

PR #316, opened 2018-08-17
I2096 - use onetime access token for setting password

littledice:
This is a breaking change, I couldn't think of a way to get around that give it involves changing the email we send out. So this should only be merged in step with the corresponding webapps PR.


src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqUserRepository.kt
```diff
@@ -202,6 +202,8 @@ class JooqUserRepository(dsl: DSLContext) : JooqRepository(dsl), UserRepository
             username = newusername
             userGroup = newusername
         }.insert()
+
+        ensureUserHasRole(newusername, ReifiedRole("user", Scope.Global()))
```
littledice: In order to issue a onetime access token that can be used to authenticate against `password/set`, the user must have the `*/can-login` permission, so we need to give them this on creation.

littledice:
Fixed tests (trailing slash again!) and added webmodels PR: https://github.com/vimc/montagu-webmodels/pull/63

littledice:
But reminder not to merge yet, webapps PR needed first

PR #319, opened 2018-08-21
I2101 - only return coverage sets metadata if user has permission

littledice:
Make scenario endpoints more permissive, but filter the output based on permissions.

Also: https://github.com/vimc/montagu-webmodels/pull/65/files


src/app/src/main/kotlin/org/vaccineimpact/api/app/logic/CoverageLogic.kt
```diff
@@ -38,7 +36,7 @@ class RepositoriesCoverageLogic(private val modellingGroupRepository: ModellingG
         // We don't use the returned responsibility, but by using this method we check that the group exists
         // and that the group is responsible for the given scenario in the given touchstoneVersion
         val responsibilityAndTouchstone = responsibilitiesRepository.getResponsibility(groupId, touchstoneVersionId, scenarioId)
-        val scenarioAndCoverageSets = touchstoneRepository.getScenario(touchstoneVersionId, scenarioId)
```
littledice: Renamed for clairty

docs/schemas/ScenarioAndCoverageSets.schema.json
```diff
@@ -9,7 +9,6 @@
 	"additionalProperties": false,
 	"required": [
 		"touchstone_version",
-		"scenario",
-		"coverage_sets"
+		"scenario"
```
MartinEden: It looks to me like below you always return the coverage sets, you just return them as null if the user doesn't have permission. So rather than making this not required, it should just be nullable?

MartinEden: Ah, I found the webmodels PR. Never mind.

PR #320, opened 2018-08-21
new methods in scenario repo

littledice:
A couple of new methods to return scenario(s) for a given touchstone version, and renaming of old methods to make clearer what they return.


src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/ScenarioRepository.kt
```diff
@@ -8,4 +9,9 @@ interface ScenarioRepository : Repository
     fun checkScenarioDescriptionExists(id: String): Unit
 
     fun getScenarios(descriptionIds: Iterable<String>): List<Scenario>
+
+    fun getScenariosForTouchstone(touchstoneVersionId: String, scenarioFilterParameters: ScenarioFilterParameters):
```
littledice: Although the existing methods for retrieving scenarios with coverage sets are in `TouchstoneRepository`, this seemed like the better place for these. Am leaving the existing methods where they are for now, but as we introduce `ScenarioLogic` and `TouchstoneLogic` classes they will probably end up being moved/refactored.

src/databaseTests/src/test/kotlin/org/vaccineimpact/api/databaseTests/tests/ScenarioTests.kt
```diff
@@ -48,6 +51,107 @@ class ScenarioTests : RepositoryTests<ScenarioRepository>()
...
     }
 
+    @Test
+    fun `getScenariosForTouchstone returns ordered scenarios`()
+    {
+        val disease1 = "disease-abc"
+        val disease2 = "disease-def"
+        withDatabase {
+            setUp(it)
+            setUpDisease(it, disease2)
```
MartinEden: I know you didn't introduce this hear, but this seems like a good time to improve the name of this function. e.g. `createDiseaseWithThreeScenarios` or something. Otherwise these tests aren't very readable.

src/databaseTests/src/test/kotlin/org/vaccineimpact/api/databaseTests/tests/ScenarioTests.kt
```diff
@@ -48,6 +51,107 @@ class ScenarioTests : RepositoryTests<ScenarioRepository>()
...
+            setUpDisease(it, disease1)
+        }
+        withRepo {
+            val scenarios = it.getScenariosForTouchstone("fake-id",
+                    ScenarioFilterParameters(disease = "fake-disease"))
+
+            assertThat(scenarios.count()).isEqualTo(0)
```
MartinEden: This test might be a bit stronger if it filtered to one of the real diseases and then checked that a) you got 3 rows b) all three scenarios belonged to right disease.

littledice:
I think the schema failure was due to a temporary connection failure, passing now.

PR #327, opened 2018-08-24
Don't return access control allow credentials for logins, or on failure

MartinEden:
This turns out not to be needed, adding a PR and closing it so we remember


MartinEden:
This doesn't make any difference to Chrome's behaviour. So I've opened https://github.com/vimc/montagu-api/pull/328 instead.

hillalex (commented):
testing webhook

PR #334, opened 2018-11-13
I2367 - validate missing rows when closing a set

littledice:
### Pre-requisites
First please review and merge https://github.com/vimc/montagu-webmodels/pull/69
Then point the webmodels submodule here back to master.

This db change must also be merged and `config/db_version` pointed back to master
https://github.com/vimc/montagu-db/pull/125

### Context 
There are 2 ways a set gets closed at the moment
1. via the `close` endpoint
2. via the `populate` endpoint when the `keepOpen` flag is missing or false

In the second case, there may also be a `redirect` parameter passed, in which case the user gets redirected back to an external url with the result encoded in a query string (this is because we use a native form post from the web app to this endpoint.)

Also relevant to note that all controller actions take place within a database transaction and if any error is thrown during the execution of that action the transaction is rolled back. In the case of the missing rows error, we don't want this to happen because we want our change to the database (marking the set as invalid) to be persisted.


src/app/src/main/kotlin/org/vaccineimpact/api/app/SequenceExtensions.kt
```diff
@@ -10,7 +10,7 @@ import org.vaccineimpact.api.models.BurdenEstimateWithRunId
 // Note that the exception will originate from whatever part of the program is currently
 // asking for another element from the sequence, not from the caller of
 // this function.
-fun Sequence<BurdenEstimateWithRunId>.validate(expectedRows: HashMap<String, HashMap<Int, HashMap<Int, Boolean>>>
+fun Sequence<BurdenEstimateWithRunId>.validate(expectedRows: HashMap<String, HashMap<Short, HashMap<Short, Boolean>>>
```
littledice: In the `expectations` table in the database ages and years are `SMALLINT`s, so I've moved to using `Short` everywhere in this code.

src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqBurdenEstimateRepository.kt
```diff
@@ -38,8 +42,42 @@ class JooqBurdenEstimateRepository(
         stochasticBurdenEstimateWriter: StochasticBurdenEstimateWriter? = null
 ) : JooqRepository(dsl), BurdenEstimateRepository
 {
-    private val centralBurdenEstimateWriter: BurdenEstimateWriter = centralBurdenEstimateWriter ?:
-            CentralBurdenEstimateWriter(dsl)
+    private fun getCountriesAsLookup(): Map<Short, String> = dsl.select(Tables.COUNTRY.ID, Tables.COUNTRY.NID)
+            .from(Tables.COUNTRY)
```
littledice: The estimates come in with a country ISO code, which is `country.id` in the db, but they are stored in the `burden_estimate` table against a numeric index which is `country.nid`. So when reading the estimates out and comparing them to the expectations (which are stored against the ISO code) we need a lookup from numeric index to ISO code, which is what this is doing.

src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqBurdenEstimateRepository.kt
```diff
@@ -38,8 +42,42 @@ class JooqBurdenEstimateRepository(
...
+                }
+
+        return expectedRowMap
+    }
+
+    private fun validate(r: BurdenEstimateRecord, countries: Map<Short, String>, expectedRows: HashMap<String, HashMap<Short, HashMap<Short, Boolean>>>)
+    {
+        val countryId = countries[r.country]
+        val ages = expectedRows[countryId]
+                ?: throw BadRequest("We are not expecting data for country $countryId")
```
littledice: Technically none of these `BadRequest`s should ever be thrown because estimates have already been validated before populating the table. I left it in for like, added safety, but perhaps we can just assume no unexpected rows?

src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqBurdenEstimateRepository.kt
```diff
@@ -268,25 +306,6 @@ class JooqBurdenEstimateRepository(
         }
     }
 
-    override fun closeBurdenEstimateSet(setId: Int, groupId: String, touchstoneVersionId: String, scenarioId: String)
-    {
-        // Dereference modelling group IDs
```
littledice: This has moved to the `BurdenEstimateLogic` class.

PR #338, opened 2018-11-29
I1287

EmmaLRussell:


src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqTouchstoneRepository.kt
```diff
@@ -170,6 +171,7 @@ class JooqTouchstoneRepository(
 
         return coverageRecords
                 .filter { it[COVERAGE.ID] != null }
+                .filter{ it[COUNTRY.NAME] != null }
```
littledice: I think I know what's going on here.  I can't see `COVERAGE.ID` being selected in the new query (which also makes sense since we're now possibly looking at multiple rows), but I think that jooq is interpreting this `ID` field then as any other field with the same name (in this case coverage_set.id) hence why this filter clause stopped working and you had to add the second filter step. 

So you can just delete the first filter step (similarly on line 176)

src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqTouchstoneRepository.kt
```diff
@@ -191,14 +193,62 @@ class JooqTouchstoneRepository(
...
+                COVERAGE.YEAR,
+                COVERAGE.COUNTRY,
+                COVERAGE.AGE_FROM,
+                COVERAGE.AGE_TO,
+                COVERAGE.AGE_RANGE_VERBATIM,
+                COVERAGE.GAVI_SUPPORT,
+                COVERAGE.GENDER)
+    }
+
+    private fun targetValid() : Field<BigDecimal?>
```
littledice: Maybe `validTargetOrNull` would be a bit clearer? Similarly for `coverageValid`

src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqTouchstoneRepository.kt
```diff
@@ -207,9 +257,13 @@ class JooqTouchstoneRepository(
                 .joinPath(COVERAGE, COUNTRY, joinType = JoinType.LEFT_OUTER_JOIN)
                 .where(TOUCHSTONE.ID.eq(touchstoneVersionId))
                 .and(SCENARIO.SCENARIO_DESCRIPTION.eq(scenarioDescriptionId))
+                .groupBy(*COVERAGE_SET.fields(),
```
littledice: Wow, had no idea about this asterisk operator, that's handy!

src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqTouchstoneRepository.kt
```diff
@@ -191,14 +193,62 @@ class JooqTouchstoneRepository(
...
+    }
+
+    private fun aggregatedCoverageValues() : List<Field<*>>
+    {
+        return arrayOf(
+                //Aggregated coverage - the sum of each row's coverage * target (to get total no of fvp's across campaign)
+                //divided by the the total target population
+                //Danger of divide by zero error here - treat as NULL if summed target is 0
+                //This is rounded to 2 dec places
+                round(
```
littledice: Turning zeros into nulls is a problem unfortunately. In the case of routine coverage (so where `COVERAGE_SET.ACTIVITY_TYPE == 'routine'`), all target values will (legitimately) be zero. This FVP calculation won't be appropriate there - we'll have to just return coverage values as is.

src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqTouchstoneRepository.kt
```diff
@@ -191,14 +191,75 @@ class JooqTouchstoneRepository(
...
+    }
+
+    private fun aggregatedValues() : List<Field<*>>
+    {
+        return arrayOf(
+                //Aggregated coverage - the sum of each row's coverage * target (to get total no of fvp's across campaign)
+                //divided by the the total target population
+
+                //This is rounded to 2 dec places if required
+                `when`(trunc(aggregatedCoverage(),2).eq(aggregatedCoverage()), aggregatedCoverage())
```
EmmaLRussell: It occurred to me that we could maybe use a user-defined function to calculate these rounded values - in a function you could store the aggregatedCoverage result in a declared variable so you wouldn't need to keep recalculating it in a single SQL expression. Not sure if we're already using UDFs elsewhere? Of course they have a maintenance overhead, and potentially a performance overhead of their own. I assume you can use a function on all the rows of a result set? You could in SQL Server. Maybe one to try if performance of this solution is poor...

src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqTouchstoneRepository.kt
```diff
@@ -191,14 +191,75 @@ class JooqTouchstoneRepository(
...
+        return `when`(COVERAGE.TARGET.isNull(), COVERAGE.TARGET) //This is actually the easiest way to get a known NULL here!
+                .otherwise(COVERAGE.COVERAGE_)
+    }
+
+    private fun aggregatedValues() : List<Field<*>>
+    {
+        return arrayOf(
+                //Aggregated coverage - the sum of each row's coverage * target (to get total no of fvp's across campaign)
+                //divided by the the total target population
+
```
EmmaLRussell: It occurred to me that we could maybe use a user-defined function to calculate these rounded values - in a function you could store the aggregatedCoverage result in a declared variable so you wouldn't need to keep recalculating it in a single SQL expression. Not sure if we're already using UDFs elsewhere? Of course they have a maintenance overhead, and potentially a performance overhead of their own. I assume you can use a function on all the rows of a result set? You could in SQL Server. Maybe one to try if performance of this solution is poor...

src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqTouchstoneRepository.kt
```diff
@@ -191,14 +191,75 @@ class JooqTouchstoneRepository(
...
+                        .otherwise(round(aggregatedCoverage(),2)).`as`("coverage"),
+                //Aggregated target
+                `when`(trunc(aggregatedTarget(),2).eq(aggregatedTarget()), aggregatedTarget())
+                        .otherwise(round(aggregatedTarget(),2)).`as`("target")
+        ).toList()
+    }
+
+    private fun aggregatedCoverage() : Field<BigDecimal?>
+    {
+        //Danger of divide by zero error here - treat as NULL if summed target is 0
```
EmmaLRussell: The rounding here was necessary because I think anytime you put in a div, Postgres increases the precision of the resulting number, even if it doesn't need to. So I was ending up with 0.2500000000000 even after my trunc comparison since 0.25 = 0.2500000000000. Probably a corner case, maybe not worth it. What this does mean is that aggregated values across groups of >1 will have a precision of 2, so might have 1 or 2 trailing zeroes (e.g. 0.70). This is probably OK since subnational is a rarity and won't affect majority of rows.

src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqTouchstoneRepository.kt
```diff
@@ -191,14 +191,80 @@ class JooqTouchstoneRepository(
...
+
+    private fun aggregatedValues() : List<Field<*>>
+    {
+        return arrayOf(
+                roundValueIfRequired(::aggregatedCoverage).`as`("coverage"),
+                roundValueIfRequired(::aggregatedTarget).`as`("target")
+        ).toList()
+    }
+
+    private fun roundValueIfRequired(valueFun : () -> Field<BigDecimal?>) : Field<BigDecimal?>
```
EmmaLRussell: It occurred to me that we could maybe use a user-defined function to calculate these rounded values - in a function you could store the aggregated value result in a declared variable so you wouldn't need to keep recalculating it in a single SQL expression. Not sure if we're already using UDFs elsewhere? Of course they have a maintenance overhead, and potentially a performance overhead of their own. I assume you can use a function on all the rows of a result set? You could in SQL Server. Maybe one to try if performance of this solution is poor...

src/blackboxTests/src/test/kotlin/org/vaccineimpact/api/blackboxTests/tests/Coverage/GroupCoverageTests.kt
```diff
@@ -99,15 +148,152 @@ class GroupCoverageTests : CoverageTests()
...
+        //5: "country_code", 6: "country", 7: "age_first", 8: "age_last", 9: "age_range_verbatim", 10: "coverage_$testYear",
+        //11: "coverage_1985", 12: "coverage_1990", 13: "coverage_1995", 14: "coverage_2000",
+        //15: "target_$testYear", 16: "target_1985", 17: "target_1990", 18: "target_1995", 19: "target_2000"
+
+        val firstRow = csv.drop(1).first().toList()
+        val expectedAggregatedTarget = "1000".toBigDecimalOrNull()
+        val expectedAggregatedCoverage = "0.9".toBigDecimalOrNull()
+
+        //test all target values
+        Assertions.assertThat(firstRow[15].toBigDecimalOrNull()).isEqualTo(expectedAggregatedTarget)
```
littledice: In this context would it make a bit more sense to just check the actual string value against what we expect, rather than casting them both using this 'BigDecimalOrNull' method? So check that e.g. `firstRow[15] == "100"` with no zeros as we expect? I think that's what we're really trying to test here. Similarly for these other tests...

src/blackboxTests/src/test/kotlin/org/vaccineimpact/api/blackboxTests/tests/Coverage/TouchstoneCoverageTests.kt
```diff
@@ -50,6 +65,114 @@ class TouchstoneCoverageTests : CoverageTests()
         test.run()
     }
 
+    @Test
```
littledice: So are there also tests for scenarios _without_ subnational rows that check the expected values are returned?

src/databaseTests/src/test/kotlin/org/vaccineimpact/api/databaseTests/tests/touchstoneRepository/GetScenarioTests.kt
```diff
@@ -213,6 +215,248 @@ class GetScenarioTests : TouchstoneRepositoryTests()
         }
     }
 
+    @Test
```
littledice: Wow, so this class is now massive and to be honest I'm finding it hard to review, so many similarly named tests!
Could we split into sub classes? Maybe leave the tests that don't actually get coverage data, like the first 4 tests in this class and create two more, `GetCoverageTests` and `GetAggregatedCoverageTests` which test the single row/grouped row logic respectively? Or any other division that makes sense to you, but I'd say this is at least 2 if not 3 classes worth of logic!

PR #340, opened 2018-12-10
I2369 - document HTTP status codes returned by Montagu

littledice:
Deleted a few unused error classes while I was at it.


src/blackboxTests/src/test/kotlin/org/vaccineimpact/api/blackboxTests/tests/BurdenEstimates/CloseBurdenEstimateSetTests.kt
```diff
@@ -109,8 +109,9 @@ class CloseBurdenEstimateSetTests : BurdenEstimateTests()
     {
         val requestHelper = RequestHelper()
 
-        val setId = JooqContext().use {
-            setUpWithBurdenEstimateSet(it)
+        val setId = 123
```
littledice: This is another attempt at stopping that non-deterministic test failure, specifying the burden estimate set id to use so it doesn't just get assigned id `1`. Still can't really explain why it's happening unfortunately.

docs/spec/spec.md
```diff
@@ -50,6 +50,21 @@ Schema: [`Response.schema.json`](../schemas/Response.schema.json)
         ]
     }
 
+# Standard response codes
```
EmmaLRussell: Might be nice to just note what each of these mean next to each number? (I couldn't remember what 409 was!) Go ahead and merge if you don't think it's worth it though!

src/blackboxTests/src/test/kotlin/org/vaccineimpact/api/blackboxTests/tests/BurdenEstimates/CloseBurdenEstimateSetTests.kt
EmmaLRussell: Oh, so it gets assigned a different set id to the other test so they don't clash? Worth putting a comment in to that effect?

PR #341, opened 2018-12-14
I1287 serializer

EmmaLRussell:
This is variation of the branch which uses the MontaguSerializer to do the rounding of excessive precision rather than the SQL query


src/serialization/src/main/kotlin/org/vaccineimpact/api/serialization/MontaguSerializer.kt
```diff
@@ -122,10 +124,18 @@ open class MontaguSerializer : Serializer
...
+
+    private fun serializeBigDecimal(value: BigDecimal) : String
+    {
+        return decimalFormat.format(value)
+    }
+
     override fun serializeValueForCSV(value: Any?) = when (value)
     {
         null -> serializeNullsTo
         is Enum<*> -> serializeEnum(value)
```
EmmaLRussell: Is checking for BigDecimal the right thing to do here, or should generalize to other floating point types we might see?

src/databaseTests/src/test/kotlin/org/vaccineimpact/api/databaseTests/tests/touchstoneRepository/GetScenarioTests.kt
```diff
@@ -12,15 +12,10 @@ import org.vaccineimpact.api.models.*
 
 class GetScenarioTests : TouchstoneRepositoryTests()
```
EmmaLRussell: This class had become a monster, so GetCoverageDataForScenarioTests and GetCoverageDataForResponsibiityTests have been split off from it

PR #346, opened 2019-01-26
Update Responsibilities burden estimate set spec

weshinsley:
Updates for the Responsibilities endpoints spec.


weshinsley:
General point added.

PR #350, opened 2019-02-13
add modelling groups to jwt token

littledice:
As well as configured permissions, all users should be granted access to their own modelling groups.

### DIFF
--- a/src/security/src/main/kotlin/org/vaccineimpact/api/security/WebTokenHelper.kt
+++ b/src/security/src/main/kotlin/org/vaccineimpact/api/security/WebTokenHelper.kt
@@ -74,8 +74,11 @@ open class WebTokenHelper(
 
     private fun modelReviewClaims(user: InternalUser): Map<String, Any>
     {
-        val modelsToReview = getReviewersMap()[user.username]?.map { it to "true" }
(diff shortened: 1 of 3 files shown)


src/security/src/main/kotlin/org/vaccineimpact/api/security/WebTokenHelper.kt
```diff
@@ -74,8 +74,11 @@ open class WebTokenHelper(
 
     private fun modelReviewClaims(user: InternalUser): Map<String, Any>
     {
-        val modelsToReview = getReviewersMap()[user.username]?.map { it to "true" }
-                ?.toMap() ?: mapOf()
+        val modelsToReview = getReviewersMap()[user.username]?: listOf()
+        val groupPermissions = (user.roles.filter { it.name == "member" }
+                .map{ it.scope.databaseScopeId } + modelsToReview)
```
EmmaLRussell: I wondered if there was any circumstances in which you might want to distinguish the groups which a user is a member of vs the groups they are a reviewer of. But there's no need for that in our current use case since both members and reviewers should be able to see all docs for their groups (if modellers wanted to keep docs private to themselves they wouldn't be sharing them with us!)

PR #355, opened 2019-02-15
Return metadata for a single burden estimate set

EmmaLRussell:


src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/BurdenEstimateRepository.kt
```diff
@@ -16,7 +16,10 @@ interface BurdenEstimateRepository : Repository
                                 properties: CreateBurdenEstimateSet,
                                 uploader: String, timestamp: Instant): Int
 
+    fun getBurdenEstimateSet(groupId: String, touchstoneVersionId: String, scenarioId: String,
```
littledice: Why not just replace the existing `getBurdenEstimateSet` method by this one? As far as I can see it's only used when populating a set, and we should be validating that the group/touchstone/scenario path is the correct one there also.

src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqBurdenEstimateRepository.kt
```diff
@@ -160,6 +159,16 @@ class JooqBurdenEstimateRepository(
                 ?: throw UnknownObjectError(setId, "burden-estimate-set")
     }
 
+    override fun getBurdenEstimateSet(groupId: String, touchstoneVersionId: String, scenarioId: String,
+                                      burdenEstimateSetId: Int): BurdenEstimateSet
+    {
+        //Get burden estimate set with validation
+        //Confirm that this burden estimate set belongs to this group, touchstone and scenario
+        checkBurdenEstimateSetPathParameters(groupId, touchstoneVersionId, scenarioId, burdenEstimateSetId)
```
littledice: Could also then just throw the 
```
                .and(RESPONSIBILITY_SET.MODELLING_GROUP.eq(groupId))
                .and(RESPONSIBILITY_SET.TOUCHSTONE.eq(touchstoneVersionId))
                .and(SCENARIO_DESCRIPTION.ID.eq(scenarioId))
```
clauses into the existing select statement, rather than doing 2 queries.

src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqBurdenEstimateRepository.kt
```diff
@@ -208,6 +217,23 @@ class JooqBurdenEstimateRepository(
...
+                .join(RESPONSIBILITY)
+                .on(RESPONSIBILITY.ID.eq(BURDEN_ESTIMATE_SET.RESPONSIBILITY))
+                .joinPath(RESPONSIBILITY, SCENARIO, SCENARIO_DESCRIPTION)
+                .joinPath(RESPONSIBILITY, RESPONSIBILITY_SET)
+                .where(BURDEN_ESTIMATE_SET.ID.eq(burdenEstimateSetId))
+                .and(RESPONSIBILITY_SET.MODELLING_GROUP.eq(groupId))
+                .and(RESPONSIBILITY_SET.TOUCHSTONE.eq(touchstoneVersionId))
+                .and(SCENARIO_DESCRIPTION.ID.eq(scenarioId))
+                .fetch()
+                .singleOrNull() ?: throw UnknownObjectError(burdenEstimateSetId, "burden estimate set id")
```
littledice: Above we use `burden-estimate-set` as the key rather than `burden-estimate-set-id` (and in general we refer to the object rather than the id.) But in fact we should really be using 
```
UnknownObjectError(burdenEstimateSetId, BurdenEstimateSet::class)
```
I know we've been using the string constructor for these errors in loads of places but ideally whenever there's a model class we should be using that rather than constructing our own string, that way we'll avoid consistency issues like this one. Going to try and be more vigilant about this!

PR #356, opened 2019-02-15
Fixed bug with parameter ordering when checking burden estimate set e…

EmmaLRussell:
…xists for responsibility

### DIFF
--- a/src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqBurdenEstimateRepository.kt
+++ b/src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqBurdenEstimateRepository.kt
@@ -396,7 +396,7 @@ class JooqBurdenEstimateRepository(
 
         // make sure set belongs to responsibility
         val responsibilityInfo = getResponsibilityInfo(modellingGroup.id, touchstoneVersionId, scenarioId)
-        val set = getBurdenEstimateSetForResponsibility(responsibilityInfo.id, setId)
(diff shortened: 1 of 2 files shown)


src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqBurdenEstimateRepository.kt
```diff
@@ -396,7 +396,7 @@ class JooqBurdenEstimateRepository(
 
         // make sure set belongs to responsibility
         val responsibilityInfo = getResponsibilityInfo(modellingGroup.id, touchstoneVersionId, scenarioId)
-        val set = getBurdenEstimateSetForResponsibility(responsibilityInfo.id, setId)
+        val set = getBurdenEstimateSetForResponsibility(setId, responsibilityInfo.id)
```
EmmaLRussell: I'm not sure this would ever have worked! We get away with it in the tests I guess because we're just creating one of everything, so everything has id 1. Is this the first time clear has been used in anger?

littledice: Ha, wow! I guess we've never used this endpoint :laughing:

weshinsley: Building the R-client is a nice way to systematically test most of the endpoints!

PR #357, opened 2019-02-19
I2652

EmmaLRussell:


src/app/src/main/kotlin/org/vaccineimpact/api/app/logic/BurdenEstimateLogic.kt
```diff
@@ -44,6 +49,60 @@ class RepositoriesBurdenEstimateLogic(private val modellingGroupRepository: Mode
...
+        val burdenOutcomeValues =
+                burdenOutcomeRows.associateBy({it.burden_outcome_code}, {it.value})
+
+        return BurdenEstimate(
+                reference.disease,
+                reference.year,
+                reference.age,
+                reference.country,
+                reference.countryName,
+                cohortSize ?: 0f,
```
EmmaLRussell: Assuming that if no cohort size has been provided, return 0

src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqBurdenEstimateRepository.kt
```diff
@@ -215,6 +217,59 @@ class JooqBurdenEstimateRepository(
...
+                )
+                .fromJoinPath(BURDEN_ESTIMATE_SET, BURDEN_ESTIMATE)
+                .join(RESPONSIBILITY)
+                .on(BURDEN_ESTIMATE_SET.RESPONSIBILITY.eq(RESPONSIBILITY.ID))
+                .joinPath(RESPONSIBILITY, SCENARIO, SCENARIO_DESCRIPTION, DISEASE)
+                .join(COUNTRY)
+                .on(BURDEN_ESTIMATE.COUNTRY.eq(COUNTRY.NID))
+                .joinPath(BURDEN_ESTIMATE, BURDEN_OUTCOME)
+                .where(BURDEN_ESTIMATE_SET.ID.eq(burdenEstimateSetId))
+                .orderBy(
```
EmmaLRussell: Not sure if this ordering is best or desirable. It makes testing easier!

hillalex: Seems reasonable enough!

src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqBurdenEstimateRepository.kt
```diff
@@ -215,6 +217,59 @@ class JooqBurdenEstimateRepository(
...
+                        DISEASE.ID,
+                        BURDEN_ESTIMATE.YEAR,
+                        BURDEN_ESTIMATE.AGE,
+                        COUNTRY.ID,
+                        COUNTRY.NAME,
+                        BURDEN_OUTCOME.CODE)
+                .fetchInto(BurdenEstimateOutcome::class.java)
+                .asSequence()
```
hillalex: I don't think there's any point calling `asSequence` here, as in the next step the sequence is evaluated anyway in order to group. But I think there's some advantage with a change to the order of operations. `fetchInto` is equivalent to one `fetch()` which returns a list followed by one `map{it.into(...)}` which also returns a list. If the order was `fetch().asSequence().map{it.into(...)}` then the last `map` operation would return a sequence, meaning one less in memory copy. Or even better `fetchSequence().map...`, then no intermediate data structures I think. 

It would be nice to have some way of benchmarking this (and other operations with large datasets, e.g.coverage data) - maybe something to think about.

Also all this reminds me that we're super close to completing this ticket, not sure why we haven't done it yet, we should bring it in to the next sprint! https://vimc.myjetbrains.com/youtrack/issue/VIMC-897

src/blackboxTests/src/test/kotlin/org/vaccineimpact/api/blackboxTests/tests/BurdenEstimates/RetrieveBurdenEstimateTests.kt
```diff
@@ -66,6 +69,311 @@ class RetrieveBurdenEstimateTests : BurdenEstimateTests()
...
+        Assertions.assertThat(dataRow[5]).isEqualTo("150.0")
+        Assertions.assertThat(dataRow[6]).isEqualTo("60.0")
+        Assertions.assertThat(dataRow[7]).isEqualTo("30.5")
+        Assertions.assertThat(dataRow[8]).isEqualTo("4.0")
+
+    }
+
+    @Test
+    fun `can get burden estimate data with some missing outcome values`()
+    {
```
hillalex: Blackbox tests take longer to run cos they're making requests to the API, so it'd generally be better to test logic like this in the Database tests.

src/databaseTests/src/test/kotlin/org/vaccineimpact/api/databaseTests/tests/burdenEstimateRepository/RetrieveBurdenEstimatesTests.kt
```diff
@@ -373,6 +375,63 @@ class RetrieveBurdenEstimatesTests : BurdenEstimateRepositoryTests()
         }
     }
 
+    @Test
+    fun `can getBurdenEstimateOutcomesSequence`()
```
hillalex: Can we add a few more tests here to test that data for the correct touchstone/scenario/set is returned? Want to avoid more bugs like that one where the responsibility set id and set id were the wrong way round :laughing:

src/databaseTests/src/test/kotlin/org/vaccineimpact/api/databaseTests/tests/burdenEstimateRepository/RetrieveBurdenEstimatesTests.kt
```diff
@@ -373,6 +375,63 @@ class RetrieveBurdenEstimatesTests : BurdenEstimateRepositoryTests()
...
+            Assertions.assertThat(result[1].countryName).isEqualTo("DEF-Name")
+            Assertions.assertThat(result[1].burden_outcome_code).isEqualTo("cohort_size")
+            Assertions.assertThat(result[1].value).isEqualTo(5f)
+
+        }
+    }
+
+    @Test
+    fun `can getExpectedOutcomesForBurdenEstimateSet`()
+    {
```
hillalex: similarly think we need a test here that only data for the desire set is returned

src/app/src/test/kotlin/org/vaccineimpact/api/tests/logic/BurdenEstimateLogicTests.kt
```diff
@@ -141,6 +142,87 @@ class BurdenEstimateLogicTests : MontaguTests()
         verify(repo).getEstimateWriter(defaultEstimateSet)
     }
 
+    @Test
+    fun `gets burden estimate set data`()
+    {
```
hillalex: Can we test the other bits of logic - that an invalid id combo throws an error, that stochastic sets are unsupported.

src/blackboxTests/src/test/kotlin/org/vaccineimpact/api/blackboxTests/tests/BurdenEstimates/RetrieveBurdenEstimateTests.kt
EmmaLRussell: Fair enough, though what these tests are really testing is the logic coordinated by the BurdenEstimateLogic class in putting the actual BurdenEstimateOutcomes together with the expectations - so I'll move these tests into the unit tests for the logic class.

hillalex: Even better, the logic test are the quickest and easiest to run :)

src/app/src/test/kotlin/org/vaccineimpact/api/tests/logic/BurdenEstimateLogicTests.kt
EmmaLRussell: Yep, putting those checks into the database tests

hillalex:
Oops, one thing missing - new endpoint needs adding to spec

PR #360, opened 2019-02-21
add new method and refactor token verification methods

hillalex:
* new method to generate upload token for burden estimate file upload
* slight refactor so that regular token verification doesn't require a `OnetimeTokenChecker` to be passed in

* return a more accurate error from token verification failures

### DIFF
--- a/src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/BurdenEstimates/BurdenEstimateUploadController.kt
+++ b/src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/BurdenEstimates/BurdenEstimateUploadController.kt
@@ -4,7 +4,7 @@ import org.vaccineimpact.api.app.ResultRedirector
 import org.vaccineimpact.api.app.asResult
 import org.vaccineimpact.api.app.context.ActionContext
 import org.vaccineimpact.api.app.context.RequestDataSource
-import org.vaccineimpact.api.app.errors.BadRequest
+import org.vaccineimpact.api.app.errors.InvalidOperationError
 import org.vaccineimpact.api.app.logic.BurdenEstimateLogic
 import org.vaccineimpact.api.app.logic.RepositoriesBurdenEstimateLogic
 import org.vaccineimpact.api.app.repositories.BurdenEstimateRepository
@@ -42,7 +42,7 @@ class BurdenEstimateUploadController(context: ActionContext,
 
         if (metadata.isStochastic())
         {
-            throw BadRequest("Stochastic estimate upload not supported")
+            throw InvalidOperationError("Stochastic estimate upload not supported")
         }
 
         return tokenHelper.generateUploadEstimatesToken(
--- a/src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/BurdenEstimates/BurdenEstimateUploadController.kt
+++ b/src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/BurdenEstimates/BurdenEstimateUploadController.kt
@@ -4,6 +4,7 @@ import org.vaccineimpact.api.app.ResultRedirector
 import org.vaccineimpact.api.app.asResult
 import org.vaccineimpact.api.app.context.ActionContext
 import org.vaccineimpact.api.app.context.RequestDataSource
+import org.vaccineimpact.api.app.errors.BadRequest
 import org.vaccineimpact.api.app.logic.BurdenEstimateLogic
 import org.vaccineimpact.api.app.logic.RepositoriesBurdenEstimateLogic
 import org.vaccineimpact.api.app.repositories.BurdenEstimateRepository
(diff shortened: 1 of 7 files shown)


src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/BurdenEstimates/BurdenEstimateUploadController.kt
```diff
@@ -4,6 +4,7 @@ import org.vaccineimpact.api.app.ResultRedirector
 import org.vaccineimpact.api.app.asResult
 import org.vaccineimpact.api.app.context.ActionContext
 import org.vaccineimpact.api.app.context.RequestDataSource
+import org.vaccineimpact.api.app.errors.BadRequest
```
EmmaLRussell: Ah, I didn't realise we had BadRequest as well as InvalidOperationError. They both return 400 status codes. I've been raising an InvalidOperationError when requesting burden estimate csv data from a stochastic set, but it would probably make more sense as a BadRequest. This is an action, so would it make more sense as an InvalidOperation?

src/app/src/test/kotlin/org/vaccineimpact/api/tests/controllers/BurdenEstimates/UploadBurdenEstimatesControllerTests.kt
```diff
@@ -1,22 +1,58 @@
...
 import org.vaccineimpact.api.app.repositories.BurdenEstimateRepository
 import org.vaccineimpact.api.app.repositories.Repositories
 import org.vaccineimpact.api.app.repositories.SimpleDataSet
 import org.vaccineimpact.api.app.requests.PostDataHelper
 import org.vaccineimpact.api.models.*
+import org.vaccineimpact.api.security.WebTokenHelper
 import org.vaccineimpact.api.tests.mocks.mockCSVPostData
 
 class UploadBurdenEstimatesControllerTests : BurdenEstimateControllerTestsBase()
 {
```
EmmaLRussell: Add test for UnknownObjectError if burden estimate set doesn't exist?

src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/BurdenEstimates/BurdenEstimateUploadController.kt
```diff
@@ -28,6 +29,31 @@ class BurdenEstimateUploadController(context: ActionContext,
...
+        val metadata = estimateRepository.getBurdenEstimateSet(path.groupId,
+                path.touchstoneVersionId,
+                path.scenarioId,
+                setId)
+
+        if (metadata.isStochastic())
+        {
+            throw BadRequest("Stochastic estimate upload not supported")
+        }
+
```
EmmaLRussell: Should there be any validation checks on the filename? Is this the filename that's going to be used on the server, or is it just the filename on the client? If the former, is that a decision the client should be making? What if there's a clash of filenames?

src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/BurdenEstimates/BurdenEstimateUploadController.kt
hillalex: Good q, not always obvious which to use, but yeah this makes sense :+1:

src/app/src/test/kotlin/org/vaccineimpact/api/tests/controllers/BurdenEstimates/UploadBurdenEstimatesControllerTests.kt
hillalex: Well that's already under test for the logic method, but added a verification that the logic method was called.

src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/BurdenEstimates/BurdenEstimateUploadController.kt
hillalex: This filename won't be used on the server, other than we will check that the filename passed when uploading matches the original filename the token was fetched for. Tbh not sure how important this is, maybe the user should just be able to request an upload token, filename agnostic, and then use it in a future request with any file?

The client can pass any filename they want, we'll actually save the file using the uid encoded in the token claim (https://github.com/vimc/montagu-api/pull/360/files#diff-142e2f35e47563007a9f8cd358d540ebR50). Tbh I guess the filename is super irrelevant, the upload token (which the client is required to pass with each chunk POST) uniquely identifies the file. Maybe we just forget the filename altogether, would make things simpler?

The only reason we might care about filename is because modellers have actually requested that we save the filenames and return them as burden estimate set metadata, for their own very basic sanity checks on what they uploaded. But for that we only have to record the filename when the upload the file, not when requesting a token.

hillalex: Oh sorry I remembered my original thinking - we don't want anyone to use the same token twice! Including the filename *somewhat* guards against this. But maybe a better way would be to store the tokens in a db (the way we do for onetime tokens) and remove them after a file has been successfully uploaded.

PR #365, opened 2019-02-25
i2628 - Throw an error for erroneous scenario id on estimate-set endpoint

hillalex:
Create logic method for getting burden estimate sets and check scenario exists before looking up estimate sets.


src/app/src/main/kotlin/org/vaccineimpact/api/app/app_start/route_config/BurdenEstimatesRouteConfig.kt
```diff
@@ -23,7 +23,7 @@ object BurdenEstimatesRouteConfig : RouteConfig
 
     override val endpoints = listOf(
             // Get and create sets
-            Endpoint("$baseUrl/", controller, "getBurdenEstimates")
+            Endpoint("$baseUrl/", controller, "getBurdenEstimateSets")
```
hillalex: Also this was badly named, so have renamed for clarity.

PR #366, opened 2019-02-26
i2705

hillalex:
Check scenario exists before querying for burden estimate set metadata. Added logic method which checks scenario exists before querying db.


hillalex:
Great spot! Will add.

PR #368, opened 2019-02-26
I2707 - Add new action to cli to add all global roles to user

hillalex:
Doh, accidentally branched off i2705 so waiting for that to be merged first (https://github.com/vimc/montagu-api/pull/366)


hillalex:
Changed my mind about this, I don't think it's optimal.

PR #370, opened 2019-02-28
I2714

EmmaLRussell:
Please check that this won't break the file upload redirects!

### DIFF
--- a/src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/BurdenEstimates/BaseBurdenEstimateController.kt
+++ b/src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/BurdenEstimates/BaseBurdenEstimateController.kt
@@ -12,6 +12,7 @@ import org.vaccineimpact.api.app.repositories.ModellingGroupRepository
 import org.vaccineimpact.api.app.repositories.Repositories
 import org.vaccineimpact.api.app.repositories.ScenarioRepository
 import org.vaccineimpact.api.app.security.checkEstimatePermissionsForTouchstoneVersion
+import org.vaccineimpact.api.app.security.getAllowableTouchstoneStatusList
 import org.vaccineimpact.api.models.Result
 
 abstract class BaseBurdenEstimateController(context: ActionContext,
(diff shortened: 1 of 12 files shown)


src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/BurdenEstimates/BaseBurdenEstimateController.kt
```diff
@@ -38,11 +40,18 @@ abstract class BaseBurdenEstimateController(context: ActionContext,
...
             context: ActionContext,
             estimateRepository: BurdenEstimateRepository,
+            groupRepository: ModellingGroupRepository,
+            scenarioRepository: ScenarioRepository,
             readEstimatesRequired: Boolean = false
     ): ResponsibilityPath
     {
         val path = ResponsibilityPath(context)
-        context.checkEstimatePermissionsForTouchstoneVersion(path.groupId, path.touchstoneVersionId, estimateRepository, readEstimatesRequired)
+        //Check that modelling group exists
```
hillalex: You may have noticed that we have `Repository` classes and a more limited number of `Logic` classes. The latter are a more recent addition to the codebase since the complexity of the logic had gotten to a point where it felt necessary, but prior to that logic was split between repositories and controllers. As such, this method predates the existence of a logic layer, but I think now rightly belongs there. We're slowly moving towards Controllers -> Logic -> Repositories.

The `BurdenEstimateLogic` class already depends on the group, scenario and burden estimate repositories, so if this method moved into there all the class signatures could remain the same, rather than having to add more dependencies to the controllers.

Digging into the `context.checkEstimatePermissions...` method shows that it uses the `estimateRepo` just to get a reference to the `TouchstoneRepo` which is super weird! The other reason we introduced a logic layer is because we ended up with circular dependencies with repositories depending on each other and we want to completely deprecate repositories ever depending on other repositories. And some of the logic depends on permissions coming from the context... shall we pair on thinking about where this logic should all rightly live and tidying it up a bit?

src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/BurdenEstimates/BurdenEstimateUploadController.kt
```diff
@@ -26,6 +28,8 @@ class BurdenEstimateUploadController(context: ActionContext,
                                      private val repositories: Repositories,
                                      private val estimatesLogic: BurdenEstimateLogic,
                                      private val estimateRepository: BurdenEstimateRepository,
+                                     private val groupRepository: ModellingGroupRepository,
```
hillalex: These extra repos are no longer needed right? Similarly in the `Upload` controller...

EmmaLRussell: Oops, yeah, forgot about that!

hillalex:
Oh my god :woman_facepalming: I just realised what the bug with the redirects is.

PR #373, opened 2019-03-05
use correct transaction scope for populating burden estimates

hillalex:
Not ready for merge, controller tests need updating.


hillalex:
Not going to fix. We will deprecate redirect logic in the next sprint.

PR #374, opened 2019-03-08
I2730

EmmaLRussell:


src/app/src/main/kotlin/org/vaccineimpact/api/app/app_start/route_config/ModellingGroupRouteConfig.kt
```diff
@@ -22,6 +22,9 @@ object ModellingGroupRouteConfig : RouteConfig
                     .json()
                     .post()
                     .secure(writePermissions),
```
EmmaLRussell: Is this a reasonable route? I can't really use /modelling-groups/memberships as it will tend to confuse the routing (it would interpret 'memberships' as a :group-id for the other route.

src/app/src/main/kotlin/org/vaccineimpact/api/app/app_start/route_config/ModellingGroupRouteConfig.kt
```diff
@@ -22,6 +22,9 @@ object ModellingGroupRouteConfig : RouteConfig
...
+            Endpoint("$baseUrl/user/memberships/", controller, "getContextUserModellingGroups")
```
EmmaLRussell: Is this a reasonable route? I can't really use /modelling-groups/memberships as it will tend to confuse the routing (it would interpret 'memberships' as a :group-id for the other route.

src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqModellingGroupRepository.kt
```diff
@@ -45,6 +45,16 @@ class JooqModellingGroupRepository(
                 .map { mapModellingGroup(it) }
     }
 
+    override fun getModellingGroups(ids: Array<String>): Iterable<ModellingGroup>
+    {
```
EmmaLRussell: Is this the right logic for what we're trying to do in getting modelling groups for a user - this is only returning 'current groups' - I assume that a member of an old group isn't automatically considered to be a member of its replacement.

src/blackboxTests/src/test/kotlin/org/vaccineimpact/api/blackboxTests/helpers/TestUserHelper.kt
```diff
@@ -48,11 +51,18 @@ class TestUserHelper(private val password: String = TestUserHelper.defaultPasswo
             }
         }
         val token = tokenFetcher.getToken(email, password)
```
EmmaLRussell: I'm not completely sure why this is clearing roles, but adding in any that you need here seems to work...

src/app/src/main/kotlin/org/vaccineimpact/api/app/app_start/route_config/ModellingGroupRouteConfig.kt
hillalex: I think either `/users/:username/modelling-groups`, `/users/current-user/modelling-groups/` or `/user/modelling-groups`. Or we could just implement something like `/user/roles/` or even just `/user/` which would return metadata about the current user, and let the front-end pull out membership roles as needed...

src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/ModellingGroupController.kt
```diff
@@ -31,6 +35,12 @@ open class ModellingGroupController(
         return modellingGroupRepository.getModellingGroupDetails(groupId)
     }
 
+    fun getContextUserModellingGroups() : List<ModellingGroup>
+    {
+        val userName = context.username!!
+        return groupLogic.getUserModellingGroups(userName)
```
hillalex: can we go with `getModellingGroupsForUser`?

src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqModellingGroupRepository.kt
hillalex: Good question. I think the reason for this `replaced_by` thing is in case the PI of a group changes, in which case we probably would want all members to get added to the new group. But it's easy to do that manually (in any case we have no mechanism for changing a group name right now, so it would be done by direct sql query, meaning members could also be updated at the same time.)

So yeah I wouldn't auto-forward memberships, but I also don't see any reason not to return all groups, i.e. drop the `and(MODELLING_GROUP.REPLACED_BY.isNull)` clause.

src/app/src/test/kotlin/org/vaccineimpact/api/tests/logic/ModellingGroupLogicTests.kt
```diff
@@ -0,0 +1,53 @@
...
+import org.vaccineimpact.api.models.Scope
+import org.vaccineimpact.api.models.permissions.ReifiedRole
+import org.vaccineimpact.api.security.InternalUser
+import org.vaccineimpact.api.security.UserProperties
+import org.vaccineimpact.api.test_helpers.MontaguTests
+
+class ModellingGroupLogicTests : MontaguTests()
+{
+    @Test
+    fun `can get responsibility with expectations`()
```
hillalex: Wrong test name ;)

src/blackboxTests/src/test/kotlin/org/vaccineimpact/api/blackboxTests/tests/ModellingGroupTests.kt
```diff
@@ -36,6 +37,65 @@ class ModellingGroupTests : DatabaseTest()
...
+    {
+        JooqContext().use {
+            it.addGroup("a", "description a")
+            it.addGroup("b", "description b")
+            it.addGroup("c", "description c")
+
+            setupTestUser()
+        }
+
+        val permissions = PermissionSet(
```
hillalex: This is actually a case where I would use the DSL, as this test doesn't currently check whether the route is secured (whereas validate.. requringPermissions... does that automatically.)

src/blackboxTests/src/test/kotlin/org/vaccineimpact/api/blackboxTests/tests/ModellingGroupTests.kt
```diff
@@ -36,6 +37,65 @@ class ModellingGroupTests : DatabaseTest()
...
+
+        assertThat(groups[0]["id"].textValue()).isEqualTo("a")
+        assertThat(groups[0]["description"].textValue()).isEqualTo("description a")
+
+        assertThat(groups[1]["id"].textValue()).isEqualTo("b")
+        assertThat(groups[1]["description"].textValue()).isEqualTo("description b")
+    }
+
+    @Test
+    fun `can get empty modelling group memberships for user with no member roles`()
```
hillalex: I think this is more appropriately a controller test.

src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqModellingGroupRepository.kt
EmmaLRussell: So if we returned all groups, that would mean in this case that a replaced group which a user belonged to would be part of the list of groups which the user of the contrib portal can see. I think this would change the current behaviour, since it's currently initially getting the group returned by getModellingGroups, which does exclude replaced groups.

src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/ModellingGroupController.kt
EmmaLRussell: Yep, and will make url /user/modelling-groups

src/blackboxTests/src/test/kotlin/org/vaccineimpact/api/blackboxTests/tests/ModellingGroupTests.kt
EmmaLRussell: Probably makes even more sense as a logic test now I've split that out

src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqModellingGroupRepository.kt
EmmaLRussell: OK, leaving this as is to avoid changing behaviour, but we should look into whether this is actually used

PR #377, opened 2019-03-19
I2770 - persist incomplete estimate sets to db

hillalex:
This brings the new endpoint up to date with the existing behaviour.

Also catches `TokenValidationException` so that an informative error can be returned to the client (otherwise a 500 will be thrown.)


src/app/src/test/kotlin/org/vaccineimpact/api/tests/controllers/BurdenEstimates/UploadBurdenEstimatesControllerTests.kt
```diff
@@ -453,10 +452,4 @@ open class UploadBurdenEstimatesControllerTests : BurdenEstimateControllerTestsB
         on { scenario } doReturn scenarioRepository
     }
 
-    private val mockStream = mock<InputStream> {
```
hillalex: Just noticed this was unused

PR #381, opened 2019-04-10
VIMC-2786 Updated new password links

EmmaLRussell:
Updated new password links to go to Montagu proxy rather than admin portal

### DIFF
--- a/src/emails/src/main/resources/templates/new-user.html
+++ b/src/emails/src/main/resources/templates/new-user.html
@@ -22,7 +22,7 @@
             <li>When you log in, Montagu sets a cookie that we can use to authenticate you across components of the Montagu platform.</li>
         </ul>
         <p><strong>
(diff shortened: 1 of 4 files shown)


EmmaLRussell:
> Looks like these emails are for a password reset so should be `reset-password`? And there should be another email somewhere for creating a new account with `new-password`?

Well new-password is the right endpoint for this purpose, but maybe this is an argument for renaming both endpoints. Currently reset-password is the page where you request the reset password link, and new-password is where that link takes you (with token) to enter your new password. 

Would 'request-password-reset' and 'reset-password' (to actually enter your new password) be clearer?

We don't have functionality to create an entirely new account with new password through the proxy. Should we?

EmmaLRussell:
Just noticed that this branch was never merged, though this ticket is in Ready to Deploy (presumably because the branch in the other repo was approved & merged)

PR #385, opened 2019-06-27
Vimc 2957 Do not round demographic data for future touchstones

EmmaLRussell:
This is a bit of a messy hack to accommodate the fact that SplitData / DataTables are reshaped from long to wide, and each object gets its own serializer on construction. Not sure if it's worth doing a more serious refactor?

### DIFF
--- a/src/databaseInterface/src/main/kotlin/org/vaccineimpact/api/db/GeneralExtensions.kt
+++ b/src/databaseInterface/src/main/kotlin/org/vaccineimpact/api/db/GeneralExtensions.kt
@@ -15,7 +15,7 @@ fun Random.nextDecimal(min: Int = 0, max: Int = 1, numberOfDecimalPlaces: Int = 
     val range = max - min
     val factor = Math.pow(10.0, numberOfDecimalPlaces.toDouble()).toInt()
     val int = this.nextInt(range * factor)
-    return BigDecimal(int) / BigDecimal(factor)
+    return BigDecimal(int).divide(BigDecimal(factor), numberOfDecimalPlaces, BigDecimal.ROUND_HALF_UP)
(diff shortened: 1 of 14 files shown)


src/databaseInterface/src/main/kotlin/org/vaccineimpact/api/db/GeneralExtensions.kt
```diff
@@ -15,7 +15,7 @@ fun Random.nextDecimal(min: Int = 0, max: Int = 1, numberOfDecimalPlaces: Int =
     val range = max - min
     val factor = Math.pow(10.0, numberOfDecimalPlaces.toDouble()).toInt()
     val int = this.nextInt(range * factor)
-    return BigDecimal(int) / BigDecimal(factor)
+    return BigDecimal(int).divide(BigDecimal(factor), numberOfDecimalPlaces, BigDecimal.ROUND_HALF_UP)
```
EmmaLRussell: This was always returning integers before, because the scale of the division wasn't being specified.

src/serialization/src/main/kotlin/org/vaccineimpact/api/serialization/DataTable.kt
```diff
@@ -10,7 +10,7 @@ import kotlin.reflect.full.declaredMemberProperties
 import kotlin.reflect.full.primaryConstructor
 
 open class DataTable<T : Any>(override val data: Sequence<T>,
-                              protected val serializer: Serializer = MontaguSerializer.instance,
+                              var serializer: Serializer = MontaguSerializer.instance,
```
hillalex: I think this can remain a `protected val`

PR #387, opened 2019-07-09
Vimc 2989 Update models endpoint to provider richer metadata

EmmaLRussell:


src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqModelRepository.kt
```diff
@@ -1,28 +1,64 @@
...
+    {
+        return dsl.select(
+                MODEL.ID,
+                MODEL.DESCRIPTION,
+                MODEL.CITATION,
+                MODEL.MODELLING_GROUP,
+                MODEL.GENDER_SPECIFIC,
+                GENDER.CODE.`as`("gender"),
+                MODEL.CURRENT_VERSION)
+                .fromJoinPath(MODEL, GENDER, joinType = JoinType.LEFT_OUTER_JOIN)
```
hillalex: I think we only want to return models for which `MODEL.IS_CURRENT` is true.

src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqModelRepository.kt
```diff
@@ -1,28 +1,64 @@
...
+    }
+
     override fun all(): List<Model>
     {
-        return dsl.fetch(MODEL).into(Model::class.java)
+        val modelRecords =  modelQuery()
```
EmmaLRussell: Actually, should this just be getting current models? this would be a change to the current behaviour, which is to get all, but do you only want to see current in the Admin screen? Make it a qs param, 'currentOnly'?

src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqModelRepository.kt
```diff
@@ -1,28 +1,64 @@
...
-        val model = dsl.select()
-                .from(MODEL)
+        val modelRecord = modelQuery()
                 .where(MODEL.ID.eq(id))
-                .fetchAny()
-
-        return model?.into(Model::class.java)
+                .singleOrNull()
                 ?: throw UnknownObjectError(id, "model_id")
+
```
hillalex: While we're touching this code do you mind updating this to be `UnknownObjectError(id, Model::class)`? :nerd_face:

src/blackboxTests/src/test/kotlin/org/vaccineimpact/api/blackboxTests/tests/ModelTests.kt
```diff
@@ -54,7 +69,45 @@ class ModelTests : DatabaseTest()
...
+                        )
+                )
+            })
+
+        }
+    }
+
+
+    @Test
+    fun `can get model without current version`()
```
hillalex: I'm also debating whether or not we should throw an `UnknownObjectError` in this case and/or exclude such models from the list coming back from `/models`. Certainly if a model is marked as current but has no current version, the modellers will not be able to upload estimates. But maybe for admin purposes it's better to see that the model is in an error state?

src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqModelRepository.kt
hillalex: Should be getting current only and I don't think any need to use a query params (I can't think of a use case where we'd want to see old models, can always add that behaviour in later if we ever need it).

EmmaLRussell: Maybe an admin might? But will change as you suggest!

src/blackboxTests/src/test/kotlin/org/vaccineimpact/api/blackboxTests/tests/ModelTests.kt
EmmaLRussell: Is that something that should be added to the admin screen then? I suppose it will be obvious from the missing metadata. Maybe it never actually happens in practice.

hillalex: I think lets not worry for now, this endpoint isn't even in use.

src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqModelRepository.kt
hillalex: I think the old ones are just there for data integrity (i.e. old estimates may have been uploaded against them) - would maybe be used in a report some time if looking at how estimates changed from prev touchstones, but not really what admins are after.

EmmaLRussell:
See models PR too: https://github.com/vimc/montagu-webmodels/pull/78

src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqModelRepository.kt
```diff
@@ -23,6 +23,7 @@ class JooqModelRepository(dsl: DSLContext) : JooqRepository(dsl), ModelRepositor
                 GENDER.CODE.`as`("gender"),
                 MODEL.CURRENT_VERSION)
                 .fromJoinPath(MODEL, GENDER, joinType = JoinType.LEFT_OUTER_JOIN)
+                .where(MODEL.IS_CURRENT.eq(true))
```
EmmaLRussell: Putting this here also means you'll get an error if you ask for a specific model which is not current

hillalex: Cool, we can always adapt this in the future if needed :+1:

hillalex:
Need to update 1 failing blackbox test

PR #389, opened 2019-07-11
Vimc 3012 Endpoint to get all expectations with model and touchstone information

EmmaLRussell:
See also models PR: https://github.com/vimc/montagu-webmodels/pull/79

### DIFF
--- a/docs/spec/Expectations.md
+++ b/docs/spec/Expectations.md
@@ -29,16 +29,6 @@ Schema: [`TouchstoneModelExpectations.schema.json`](../schemas/TouchstoneModelEx
                     "minimum_birth_year": 1950,
                     "maximum_birth_year": 2050
                 },
-                "countries": [
-                    {
-                        "id": "AFG",
-                        "name": "Afghanistan"
-                    },
-                    {
-                        "id": "AGO",
-                        "name": "Angola"
-                    }
-                ],
                 "outcomes": ["deaths", "cases"]
             }
         },
@@ -61,16 +51,6 @@ Schema: [`TouchstoneModelExpectations.schema.json`](../schemas/TouchstoneModelEx
                     "minimum_birth_year": 1960,
                     "maximum_birth_year": 2060
                 },
-                "countries": [
-                    {
-                        "id": "AFG",
-                        "name": "Afghanistan"
-                    },
-                    {
-                        "id": "AGO",
-                        "name": "Angola"
-                    }
-                ],
                 "outcomes": ["cases"]
             }
         }
--- a/docs/spec/Expectations.md
+++ b/docs/spec/Expectations.md
@@ -29,7 +29,7 @@ Schema: [`TouchstoneModelExpectations.schema.json`](../schemas/TouchstoneModelEx
                     "minimum_birth_year": 1950,
                     "maximum_birth_year": 2050
                 },
-                "outcomes": [""cases","deaths"]
+                "outcomes": ["cases","deaths"]
             }
         },
         {
--- a/docs/spec/Expectations.md
+++ b/docs/spec/Expectations.md
@@ -1,0 +1,77 @@
+# Expectations
+## GET /expectations/
+Returns an enumeration of all expectations, along with the touchstone version id, modelling group and disease to which
+they apply. Expectations are returned only for open touchstones and responsibilities.
+
+Required permissions: */responsibilities.read
+
+Schema: [`TouchstoneModelExpectations.schema.json`](../schemas/TouchstoneModelExpectations.schema.json)
+
+#### Example
+
+    [
+        {
+            "touchstone_version": "2017-op-1",
+            "modelling_group": "IC-Garkse",
+            "disease": "YF",
+            "expectations": {
+                "id": 12,
+                "description": "Expectations for pilot run",
+                "years": {
+                    "minimum_inclusive": 2000,
+                    "maximum_inclusive": 2100
+                },
+                "ages": {
+                    "minimum_inclusive": 0,
+                    "maximum_inclusive": 99
+                },
+                "cohorts": {
+                    "minimum_birth_year": 1950,
+                    "maximum_birth_year": 2050
+                },
+                "countries": [
+                    {
+                        "id": "AFG",
+                        "name": "Afghanistan"
+                    },
+                    {
+                        "id": "AGO",
+                        "name": "Angola"
+                    }
+                ],
+                "outcomes": ["deaths", "cases"]
+            }
+        },
+        {
+            "touchstone_version": "2018-op-1",
+            "modelling_group": "IC-Garkse",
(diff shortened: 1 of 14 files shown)


src/app/src/main/kotlin/org/vaccineimpact/api/app/app_start/route_config/ExpectationsRouteConfig.kt
```diff
@@ -0,0 +1,20 @@
+package org.vaccineimpact.api.app.app_start.route_config
+
+import org.vaccineimpact.api.app.app_start.*
+import org.vaccineimpact.api.app.controllers.ExpectationsController
+
+object ExpectationsRouteConfig : RouteConfig
+{
+    private val getAllPermissions = setOf(
+        "*/responsibilities.read"
```
EmmaLRussell: is this the right permission to require?

src/app/src/test/kotlin/org/vaccineimpact/api/tests/controllers/ExpectationsControllerTests.kt
```diff
@@ -0,0 +1,46 @@
...
+import org.vaccineimpact.api.models.Country
+import org.vaccineimpact.api.models.TouchstoneModelExpectations
+import org.vaccineimpact.api.models.Expectations
+import org.vaccineimpact.api.test_helpers.MontaguTests
+
+
+class ExpectationsControllerTests : MontaguTests()
+{
+    @Test
+    fun `getModels returns all models`()
```
hillalex: test name!

src/databaseTests/src/test/kotlin/org/vaccineimpact/api/databaseTests/tests/ExpectationsRepositoryTests.kt
```diff
@@ -249,6 +251,90 @@ class ExpectationsRepositoryTests : RepositoryTests<ExpectationsRepository>()
...
+                    TouchstoneModelExpectations(touchstoneVersionId, groupId, "YF",
+                            exampleExpectations()),
+                    TouchstoneModelExpectations("touchstone2-2", otherGroupId, "HepB",
+                            exampleExpectations(id=2))
+            ))
+        }
+    }
+
+    @Test
+    fun `get all expectations does not return those from closed touchstones`()
```
hillalex: should also check `in-preparation` touchstones

src/app/src/main/kotlin/org/vaccineimpact/api/app/app_start/route_config/ExpectationsRouteConfig.kt
hillalex: This seems right to me

src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqExpectationsRepository.kt
```diff
@@ -87,6 +84,28 @@ class JooqExpectationsRepository(dsl: DSLContext)
                 }
     }
 
+    override fun getAllExpectations(): List<TouchstoneModelExpectations>
+    {
+        val records = dsl.select(
+                SCENARIO_DESCRIPTION.DISEASE,
+                TOUCHSTONE.ID,
+                RESPONSIBILITY_SET.MODELLING_GROUP,
+                *BURDEN_ESTIMATE_EXPECTATION.fields())
```
hillalex: So this is missing outcomes and countries. The latter aren't actually needed for this front-end work, since we're looking instead at max applicable countries, but outcomes definitely are.

docs/spec/Expectations.md
```diff
@@ -0,0 +1,77 @@
...
+                },
+                "ages": {
+                    "minimum_inclusive": 0,
+                    "maximum_inclusive": 99
+                },
+                "cohorts": {
+                    "minimum_birth_year": 1960,
+                    "maximum_birth_year": 2060
+                },
+                "countries": [
```
hillalex: Given this implementation countries and outcomes will always be empty. We could exclude countries if we want, since not needed for the admin portal work, but in that case we shouldn't return the field at all (an empty array will be misleading!)

src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqExpectationsRepository.kt
EmmaLRussell: There's a subsequent call in the map to withCountriesAndOutcomes, which looks like it should get both (though I should have added a test that it actually does!)
But if we don't need countries, I'll change that to just get outcomes (and add a test)

src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqExpectationsRepository.kt
```diff
@@ -87,6 +85,40 @@ class JooqExpectationsRepository(dsl: DSLContext)
...
+                .fromJoinPath(RESPONSIBILITY, BURDEN_ESTIMATE_EXPECTATION)
+                .joinPath(RESPONSIBILITY, SCENARIO, SCENARIO_DESCRIPTION)
+                .joinPath(RESPONSIBILITY, RESPONSIBILITY_SET, TOUCHSTONE)
+                .joinPath(BURDEN_ESTIMATE_EXPECTATION, Tables.outcomes, joinType = JoinType.LEFT_OUTER_JOIN)
+                .where(RESPONSIBILITY.IS_OPEN.eq(true))
+                .and(TOUCHSTONE.STATUS.eq("open"))
+                .and(SCENARIO.TOUCHSTONE.eq(RESPONSIBILITY_SET.TOUCHSTONE))
+                .orderBy(RESPONSIBILITY_SET.MODELLING_GROUP, SCENARIO_DESCRIPTION.DISEASE, TOUCHSTONE.ID)
+
+        val outcomes = records.groupingBy{ it[BURDEN_ESTIMATE_EXPECTATION.ID] }
```
EmmaLRussell: So this quite ugly bit of code was an attempt to get all the outcomes along with the estimates stuff, and save doing separate queries to get each of the outcomes sets (what withOutcomesAndCountries would do, I think). Not sure if there's a nicer approach?

hillalex: Wow, that is quite something, you've definitely discovered new Kotlin syntax that I've never seen before!
But I think this is clearer:

```
        val outcomes = records.groupBy { it[BURDEN_ESTIMATE_EXPECTATION.ID] }
                .mapValues {
                    it.value.mapNotNull { row -> row[Tables.outcomes.OUTCOME] }
                }
```

PR #395, opened 2019-08-06
fix missing rows message

hillalex:
Turns out this was very badly tested.

### DIFF
--- a/src/app/src/main/kotlin/org/vaccineimpact/api/app/logic/BurdenEstimateLogic.kt
+++ b/src/app/src/main/kotlin/org/vaccineimpact/api/app/logic/BurdenEstimateLogic.kt
@@ -44,6 +44,7 @@ class RepositoriesBurdenEstimateLogic(private val modellingGroupRepository: Mode
                                       private val scenarioRepository: ScenarioRepository,
                                       private val touchstoneRepository: TouchstoneRepository) : BurdenEstimateLogic
 {
+
     override fun getBurdenEstimateSets(groupId: String, touchstoneVersionId: String, scenarioId: String): List<BurdenEstimateSet>
     {
         scenarioRepository.checkScenarioDescriptionExists(scenarioId)
@@ -144,9 +145,9 @@ class RepositoriesBurdenEstimateLogic(private val modellingGroupRepository: Mode
         else
         {
             val expectedRows = expectationsRepository.getExpectationsForResponsibility(responsibilityInfo.id)
-                    .expectation.expectedRowHashMap()
+                    .expectation.expectedRowLookup()
             val validatedRowMap = burdenEstimateRepository.validateEstimates(set, expectedRows)
-            val countriesWithMissingRows = validatedRowMap.filter(::hasMissingRows)
+            val countriesWithMissingRows = validatedRowMap.filter{ it.hasMissingAges() }
             if (countriesWithMissingRows.any())
             {
                 burdenEstimateRepository.changeBurdenEstimateStatus(setId, BurdenEstimateSetStatus.INVALID)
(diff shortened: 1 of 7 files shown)


src/app/src/main/kotlin/org/vaccineimpact/api/app/logic/BurdenEstimateLogic.kt
```diff
@@ -172,26 +169,40 @@ class RepositoriesBurdenEstimateLogic(private val modellingGroupRepository: Mode
...
+
+    private fun isFalse(mapEntry: Map.Entry<Short, Boolean>): Boolean
+    {
+        return !mapEntry.value
+    }
+
+    private fun hasFalseEntry(mapEntry: Map.Entry<Short, HashMap<Short, Boolean>>): Boolean
     {
-        val countries = missingRows.keys
-        val message = "Missing rows for ${countries.joinToString(", ")}"
```
hillalex: This method was the problem. The `missingRows` argument is not actually a map containing only missing rows, rather it is a collection of all rows for countries that have at least one missing row. We can't assume that the first country/year/age combo is actually missing, we need to check for false values.

I've changed the method to do just that, changing the name of the `missingRows` argument as well to clarify. The other way to fix this would be to actually filter the whole nested hashmap to be only those rows which are missing before we pass it in to this method. But I think that would require more enumeration and result in more new objects being created (so be less performant). I haven't actually tested that though.

src/databaseTests/src/test/kotlin/org/vaccineimpact/api/databaseTests/tests/burdenEstimateRepository/PopulateBurdenEstimateSetTests.kt
```diff
@@ -99,4 +98,59 @@ class PopulateBurdenEstimateSetTests : BurdenEstimateRepositoryTests()
 
     }
 
+    @Test
+    fun `validateEstimates returns a map with missing rows`()
```
hillalex: This bit actually wasn't under test at all!

src/app/src/main/kotlin/org/vaccineimpact/api/app/logic/BurdenEstimateLogic.kt
```diff
@@ -172,26 +169,38 @@ class RepositoriesBurdenEstimateLogic(private val modellingGroupRepository: Mode
...
         scenarioRepository.checkScenarioDescriptionExists(path.scenarioId)
     }
 
-    private fun missingRows(countryMapEntry: Map.Entry<String, HashMap<Short, HashMap<Short, Boolean>>>): Boolean
+    private fun rowErrorMessage(countriesWithMissingRows: Map<String, HashMap<Short, HashMap<Short, Boolean>>>): String
     {
-        return countryMapEntry.value.any { a ->
-            a.value.any { y -> !y.value }
-        }
+        val countries = countriesWithMissingRows.keys
```
EmmaLRussell: In general, the whole way the logic code around missing rows deals with Map<String, HashMap<Short, HashMap<Short, Boolean>>> and other such opaque structures is pretty confusing. Laying the code out like this with clear val names, does really help understanding, though as you'll see I still managed to get confused about the test! It would probably be even clearer to write a class or two with obvious property/method names rather than using lengthy generics - up to you whether you think it's worth refactoring though.

hillalex: Agreed. I've added some type aliases and extension methods which I hope makes things a little clearer at the point of use.

hillalex:
Associated webmodels PR https://github.com/vimc/montagu-webmodels/pull/83

PR #397, opened 2019-08-07
Add scenarios to TouchstoneModelExpectations

EmmaLRussell:
Another pair of preliminary PRs, to add scenarios to the /expectations endpoint. 
Here's the other one: https://github.com/vimc/montagu-webmodels/pull/84

### DIFF
--- a/src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqExpectationsRepository.kt
+++ b/src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqExpectationsRepository.kt
@@ -137,7 +137,7 @@ class JooqExpectationsRepository(dsl: DSLContext)
         val scenarios = records.map { it[SCENARIO_DESCRIPTION.ID] }.distinct().sorted()
         val diseases = records.map { it[SCENARIO_DESCRIPTION.DISEASE] }.distinct()
         val disease = diseases.singleOrNull()
-                ?: throw DatabaseContentsError("Expectations $expectationsId is used by responsibilities that do not all share the same disease: ${diseases.joinToString()}")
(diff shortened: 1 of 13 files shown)


src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqExpectationsRepository.kt
```diff
@@ -104,17 +105,24 @@ class JooqExpectationsRepository(dsl: DSLContext)
 
         val outcomes = records.groupBy{ it[BURDEN_ESTIMATE_EXPECTATION.ID] }
             .mapValues{
-                it.value.mapNotNull{ row -> row[Tables.outcomes.OUTCOME] }
+                it.value.mapNotNull{ row -> row[Tables.outcomes.OUTCOME] }.distinct()
```
EmmaLRussell: Added deduping for both outcomes and scenarios and we have rows for each combination of outcome and scenario..

EmmaLRussell:
> I'm a bit concerned that the proliferation of expectation related models is getting confusing. We have `Expectations`, `OutcomeExpectations` which has the same fields as `Expectations` minus the countries, `ExpectationMapping` and `TouchstoneModelExpectations` which has the same fields as `ExpectationMapping` plus touchstone and modelling group, except for the fact that the expectations are `OutcomeExpectations` and not full `Expectations`! I can't help but feel there is a more sensible way to organise things. E.g `Expectations` could be called `CountryOutcomeExpectations` and could implement a shared interface with `OutcomeExpectations`? Then `ExpectationMapping` could be a shared interface also, with `TouchstoneModelExpectationMapping` having some extra properties?
> 
> Maybe a more radical reorganisation is warranted but I think at least some shared interfaces would help...

Yeah, we're not using inheritance in the classes themselves as data classes and inheritance don't really mix, but yes could certainly put some shared interfaced on to make clear what the differences are, maybe with some brief comments as to to reasons/usage of each.

PR #398, opened 2019-08-08
do not update current estimate set on creation

hillalex:
Should only be merged after https://github.com/vimc/montagu-webapps/pull/399


hillalex:
Closing as wrong branch name! Replaced with https://github.com/vimc/montagu-api/pull/400

PR #402, opened 2019-08-12
VIMC-3062 preliminary refactor

hillalex:
I figured while I'm touching the burden estimate set creation logic I should move it into the logic layer as per our incremental refactor. So this PR keeps functionality the same but moves logic into the logic layer.

Contains commits from https://github.com/vimc/montagu-api/pull/401 so merge that one first and merge master in here for clean diff.


EmmaLRussell:
Looks sensible to me -haven't merged as file conflicts..

PR #406, opened 2019-08-15
VIMC-2986 send Slack notifications

hillalex:
Hard (impossible?) to write automated integration tests for this but can be confirmed working by putting the correct config value in locally and running a Blackbox test that populates estimates  https://vimc.slack.com/archives/CMDP3V361/p1565889533000500

PR to inject the real webhook url during deployment: https://github.com/vimc/montagu/pull/161

### DIFF
--- a/src/app/src/main/kotlin/org/vaccineimpact/api/app/logic/BurdenEstimateLogic.kt
+++ b/src/app/src/main/kotlin/org/vaccineimpact/api/app/logic/BurdenEstimateLogic.kt
@@ -54,14 +54,17 @@ class RepositoriesBurdenEstimateLogic(private val modellingGroupRepository: Mode
                                       private val burdenEstimateRepository: BurdenEstimateRepository,
                                       private val expectationsRepository: ExpectationsRepository,
                                       private val scenarioRepository: ScenarioRepository,
-                                      private val touchstoneRepository: TouchstoneRepository) : BurdenEstimateLogic
+                                      private val touchstoneRepository: TouchstoneRepository,
+                                      private val notifier: Notifier) : BurdenEstimateLogic
 {
 
-    constructor(repositories: Repositories) : this(repositories.modellingGroup,
+    constructor(repositories: Repositories, notifier: Notifier = SlackNotifier()) : this(
+            repositories.modellingGroup,
             repositories.burdenEstimates,
             repositories.expectations,
             repositories.scenario,
-            repositories.touchstone)
+            repositories.touchstone,
+            notifier)
 
     override fun createBurdenEstimateSet(groupId: String, touchstoneVersionId: String, scenarioId: String,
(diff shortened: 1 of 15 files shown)


src/app/src/test/kotlin/org/vaccineimpact/api/tests/logic/BurdenEstimateLogic/BaseBurdenEstimateLogicTests.kt
```diff
@@ -0,0 +1,86 @@
+package org.vaccineimpact.api.tests.logic.BurdenEstimateLogic
```
hillalex: The BurdenEstimatesLogicTests class was super long so I split it up further and introduced this base class for shared logic

src/app/src/test/kotlin/org/vaccineimpact/api/tests/logic/BurdenEstimateLogic/PopulateBurdenEstimateLogicTests.kt
```diff
@@ -0,0 +1,126 @@
...
+    {
+        val repo = mockEstimatesRepository()
+        val sut = RepositoriesBurdenEstimateLogic(mockGroupRepository(), repo, mockExpectationsRepository(), mock(), mock(), mock())
+
+        sut.populateBurdenEstimateSet(setId, groupId, touchstoneVersionId, scenarioId, validData, null)
+        verify(repo).getEstimateWriter(defaultEstimateSet)
+    }
+
+    @Test
+    fun `notifies when a set has been populated`()
```
hillalex: This is the only new test. The others are just copied over.

src/app/src/main/kotlin/org/vaccineimpact/api/app/logic/BurdenEstimateLogic.kt
```diff
@@ -284,6 +287,8 @@ class RepositoriesBurdenEstimateLogic(private val modellingGroupRepository: Mode
         {
             burdenEstimateRepository.updateBurdenEstimateSetFilename(setId, filename)
         }
+
+        notifier.notify("A burden estimate set has just been uploaded for ${group.id}")
```
EmmaLRussell: Would it be worth putting more information in the message about the burden estimate? Since Kim doesn't have visibility on Touchstone progress yet, maybe scenario would be a useful thing to surface here (would at least show if multiple uploads are multiple attempts at the same set, or different scenarios). Touchstone is probably limited usefulness since I think there's only usually one active one, but could possibly be of interest. 
Also, would it be worth showing if the upload was valid or invalid?

hillalex: Both great points

PR #409, opened 2019-10-30
vimc-3149 Include gender in coverage for 2019 touchstones onwards

EmmaLRussell:
This would have been a massive PR, but it didn't really make sense to split up the feature, so I've split the PR by moving most of the tests, particularly the blackbox and PR tests into a separate PR here: https://github.com/vimc/montagu-api/pull/410
This PR should be merged first.


src/app/src/main/kotlin/org/vaccineimpact/api/app/Helpers.kt
```diff
@@ -63,4 +66,30 @@ fun Future<Exception?>.awaitAndThrowIfError()
     {
         throw exception
     }
+}
+
+@Suppress("UNCHECKED_CAST")
```
EmmaLRussell: These are here because I found DataTable.new was raising a 'no primary constructor' error if i didn't cast data explicitly now there are a choice of types of wide and long rows. Not sure if there's a better approach..

PR #410, opened 2019-11-04
vimc-3149_tests Gender in coverage with tests

EmmaLRussell:
All the missing tests, including black box tests and database tests from https://github.com/vimc/montagu-api/pull/409 which should be merged first


src/databaseTests/src/test/kotlin/org/vaccineimpact/api/databaseTests/tests/touchstoneRepository/GetCoverageDataForResponsibilityTests.kt
```diff
@@ -37,26 +37,29 @@ class GetCoverageDataForResponsibilityTests : TouchstoneRepositoryTests()
                     scenarioId)
 
             assertThat(result.toList()).containsExactlyElementsOf(listOf(
-                    LongCoverageRow(scenarioId, "First", "AF", GAVISupportLevel.WITHOUT, ActivityType.ROUTINE,
-                            "AAA", "AAA-Name", 2001, 2.toDecimal(), 4.toDecimal(), null, null, null),
+                    GenderedLongCoverageRow(scenarioId, "First", "AF", GAVISupportLevel.WITHOUT, ActivityType.ROUTINE,
```
hillalex: So `touchstoneVersionId` here is `touchstone-1`, as declared in the base class - is it clear that we should be returning gender for touchstone ids without an explicit date in the id? 

In any case could we move the declaration of `touchstoneVersionId` into the test itself, so that it's easy to read this test and see that it's returning the appropriate data given the `touchstoneVersionId`? Or could create a shared variable called `2019touchstoneVersionId` to make it clear. Would also be good to have a test with an id like `2020-somethingorother`.

EmmaLRussell:
> Re the separate tests branch, I don't think that will work as intended because it looks like `vimc-3149` and this branch share commits, but the former has some extra commits deleting code. If they both get merged, the tests will remain deleted.

Ugh, yes of course. :woman_facepalming:

src/databaseTests/src/test/kotlin/org/vaccineimpact/api/databaseTests/tests/touchstoneRepository/GetCoverageDataForResponsibilityTests.kt
EmmaLRussell: I'm not sure we really have clearly defined requirements for non-standard touchstone names, but I guess the new behaviour should be the default. Anyway, I'll change this to use a 2019 touchstone Id, and also 2020.

hillalex: Cool, just had a quick check and there are no touchstones in the db that don't start with a year :+1:

EmmaLRussell:
Webmodels PR:
https://github.com/vimc/montagu-webmodels/pull/88

hillalex:
Looks good, thanks. Want to point webmodels back to master and then merge?

PR #412, opened 2020-01-06
update kotlin to 1.3

hillalex:
Just a couple of deprecated methods in the models project: https://github.com/vimc/montagu-webmodels/pull/89


src/blackboxTests/src/test/kotlin/org/vaccineimpact/api/blackboxTests/tests/BurdenEstimates/PopulateBurdenEstimateTests.kt
```diff
@@ -373,7 +373,6 @@ class PopulateBurdenEstimateTests : BurdenEstimateTests()
     }
 
     private fun populateFromFile(setId: Int,
-                                 total: Int,
```
hillalex: Not related to the update, but compiler was warning about this unused argument.

PR #413, opened 2020-01-06
VIMC-3329 generate auth profile from db

hillalex:
Adds `MontaguAuthorizationGenerator` which uses existing user repo method to grab user data and store it in the profile.

Contains commits from https://github.com/vimc/montagu-api/pull/412


src/app/src/main/kotlin/org/vaccineimpact/api/app/security/TokenVerifyingConfigFactory.kt
```diff
@@ -29,7 +27,7 @@ class TokenVerifyingConfigFactory(
     override fun build(vararg parameters: Any?): Config
     {
         clients.forEach {
-            it.addAuthorizationGenerator({ _, profile -> extractPermissionsFromToken(profile) })
+            it.addAuthorizationGenerator(MontaguAuthorizationGenerator(repositoryFactory))
```
EmmaLRussell: Nice, this is a lot more readable than the old approach, as well as fixing the token length issue!

EmmaLRussell: Is there a missing test that this factory class is adding the expected authorization generators? (Missing anyway, before this behaviour change..)

hillalex: Overall the API test coverage isn't looking great, it's coming back at only around 30% coverage! But yeah this seems like a good point to add in a test for this class.

PR #414, opened 2020-01-06
VIMC-3332

hillalex:
Seemed like it would be easiest to add a Travis build, rather than grapple with the issue of authenticating TC agents with codecov! This does mean there's a bit of duplication across the 2 build servers - the Travis build runs the app tests and database tests to get code coverage (but not the blackbox tests, which wouldn't generate code coverage.) 

Discovered Libsodium was out of date, so have updated that to get things running on Travis.


travis/start-database.sh
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env bash
```
hillalex: This is a bit annoying - we can't re-use the existing `startDatabase` gradle task/existing scripts because they use the local registry which Travis can't access. 

We can't just switch over to using the public docker hub everywhere because that would mean we could no longer start the database with arbitrary branch or commit tag as per the `db_version` config file (because we only push master and semantic version tags to docker hub.) It feels ok to always use master here because we're just trying to get at code coverage, and tests will be run against the configured version on TeamCity.

So options:
1. maintain 2 ways of starting the database, this script which runs the public master image, and the gradle startDatabase task which runs the configured version image from the local registry
2. push all images to docker hub and stop using our local registry, change the existing task to use the public docker hub and use that everywhere

src/databaseTests/build.gradle
```diff
@@ -19,8 +19,8 @@ dependencies {
 }
 
 test {
-    dependsOn 'copySpec', ':startDatabase', 'createTemplateFromDatabase'
```
hillalex: These tasks will fail on Travis because they require the local registry, so have stopped them being triggered automatically and now manually executing them before/after tests run on TC https://github.com/vimc/montagu-api/pull/414/files#diff-3e0e48b56ab9b06df6f4e2465783c472R13

travis/start-database.sh
EmmaLRussell: Agreed. Maybe put a comment in here, or rename the file somehow to remind ourselves of why we have two mechanisms (I will definitely forget).

hillalex: Cool, yeah I agree that maintaining the 2 files is probably the better option :+1:

hillalex:
Ok, so we're up to 72% code coverage, in actual fact it will be higher but it's going to be a faff to get blackbox tests running on Travis so I'm opening a fresh ticket for that: https://mrc-ide.myjetbrains.com/youtrack/issue/VIMC-3334

PR #415, opened 2020-01-09
VIMC-3330 remove roles and permissions from tokens

hillalex:


hillalex:
> This looks find to me, but wasn't there an issue with the front end reading permissions from the token? Did that already get taken out?

I *think* that it's covered by this? https://github.com/vimc/montagu-webapps/pull/388/files
But yes let me run up the webapps with this branch of the API to confirm.

hillalex:
Yep, all seems to run fine :+1:

PR #416, opened 2020-01-14
vimc-3230 Add OrderlyWeb user on new Montagu user

EmmaLRussell:
I think this should be ready to review finally, but see issue with updating the webapps to use this:
https://github.com/vimc/montagu-webapps/pull/423
Integration test is failing on TeamCity, though add user manually works


README.md
```diff
@@ -11,9 +11,10 @@ Install Docker and add your user to the Docker group (e.g. https://www.digitaloc
 Configure your Docker client to use our registry by following instructions here:
 https://github.com/vimc/montagu-ci#configuring-docker-clients-to-use-the-registry
 
-Start the database by navigating to the `src` folder and running
+Start the dependencies (the database and orderly web) by navigating to the `src` folder and running
```
EmmaLRussell: I originally tried to put both the database start and orderly web start into a single gradle task startDependencies, but this caused a big mess on TeamCity because the build environment which runs the unit tests was running that task within a docker container, which was attempting to start the orderly containers-within-a-container with a mounted volume with a path inside the first container.

scripts/run-blackbox-tests.sh
```diff
@@ -26,7 +31,27 @@ docker exec montagu_db_1 psql -U vimc -d montagu -c \
        "CREATE USER MAPPING FOR vimc SERVER montagu_db_annex OPTIONS (user 'vimc', password 'changeme');"
 # -------------------------------------------------------------
 
-# Build and image that can run blackbox tests
+
+ORDERLY_IMAGE="vimc/orderly:master"
```
hillalex: It looks like almost all this logic this is duplicated from ./scripts/start-orderly-web - can that script be re-used here?

scripts/run-blackbox-tests.sh
```diff
@@ -26,7 +31,27 @@ docker exec montagu_db_1 psql -U vimc -d montagu -c \
...
+OW_MIGRATE_IMAGE="vimc/orderlyweb-migrate:master"
+OW_CLI_IMAGE="vimc/orderly-web-user-cli:master"
+
+# create orderly db
```
hillalex: In TC we could add the demo artifact from the orderly build as a dependency to save having to regenerate it each time (this is what we do in the OW build, and now in the webapps)

scripts/run-blackbox-tests.sh
EmmaLRussell: They're a little bit different because in the blackbox tests, orderlyweb is getting pulled up as part of the docker compose configuration, but I'm pulling out the common stuff of orderlyweb dependencies into a shared script/

hillalex:
This looks much better, thanks. I have a couple of suggestions for further tidying up the scripting: https://github.com/vimc/montagu-api/pull/417

If you're happy with those then go ahead and merge that and then this PR!

PR #418, opened 2020-02-10
vimc-3289 upload race condition bug

hillalex:
I identified the race condition as happening at the start of file uploads. Once the cache entry for a particular file exists, recording new chunks is thread safe. But *creating* the cache entry in the first place is not thread safe, and the first 3 simultaneous chunk uploads may overwrite one another.

This PR makes the race condiiton tighter. A full solution would be to have an initial endpoint that just sets up the cache entry, before any chunks are posted.

### DIFF
--- a/src/app/src/main/kotlin/org/vaccineimpact/api/app/ChunkedFileCache.kt
+++ b/src/app/src/main/kotlin/org/vaccineimpact/api/app/ChunkedFileCache.kt
@@ -5,13 +5,14 @@ import org.vaccineimpact.api.app.models.ChunkedFile
 import java.util.concurrent.ConcurrentHashMap
 import java.util.concurrent.TimeUnit
 
-interface Cache<T> {
+interface Cache<T>
+{
     operator fun get(uniqueIdentifier: String): T?
     fun put(item: T)
     fun remove(uniqueIdentifier: String)
 }
 
-class ChunkedFileCache(private val flushInterval: Long = TimeUnit.HOURS.toMillis(1)): Cache<ChunkedFile>
+class ChunkedFileCache(private val flushInterval: Long = TimeUnit.HOURS.toMillis(1)) : Cache<ChunkedFile>
 {
     private val memoryCache = ConcurrentHashMap<String, MutablePair<ChunkedFile, Long>>()
 
@@ -29,7 +30,14 @@ class ChunkedFileCache(private val flushInterval: Long = TimeUnit.HOURS.toMillis
     override fun put(item: ChunkedFile)
     {
         recycle()
-        memoryCache[item.uniqueIdentifier] = MutablePair(item, System.currentTimeMillis())
+        if (memoryCache[item.uniqueIdentifier] == null)
+        {
+            memoryCache[item.uniqueIdentifier] = MutablePair(item, System.currentTimeMillis())
+        }
+        else
+        {
+            memoryCache[item.uniqueIdentifier]!!.left.uploadedChunks.putAll(item.uploadedChunks)
+        }
     }
 
     override fun remove(uniqueIdentifier: String)
@@ -39,7 +47,8 @@ class ChunkedFileCache(private val flushInterval: Long = TimeUnit.HOURS.toMillis
         item?.left?.cleanUp()
     }
 
-    private fun recycle() {
+    private fun recycle()
+    {
         memoryCache.filter {
             it.value.right < System.currentTimeMillis() - flushInterval
         }.map {
--- a/src/app/src/main/kotlin/org/vaccineimpact/api/app/ChunkedFileCache.kt
+++ b/src/app/src/main/kotlin/org/vaccineimpact/api/app/ChunkedFileCache.kt
@@ -34,10 +34,6 @@ class ChunkedFileCache(private val flushInterval: Long = TimeUnit.HOURS.toMillis
         {
             memoryCache[item.uniqueIdentifier] = MutablePair(item, System.currentTimeMillis())
         }
-        else
-        {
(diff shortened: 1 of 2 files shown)


src/app/src/main/kotlin/org/vaccineimpact/api/app/ChunkedFileCache.kt
```diff
@@ -29,7 +30,14 @@ class ChunkedFileCache(private val flushInterval: Long = TimeUnit.HOURS.toMillis
     override fun put(item: ChunkedFile)
     {
         recycle()
-        memoryCache[item.uniqueIdentifier] = MutablePair(item, System.currentTimeMillis())
+        if (memoryCache[item.uniqueIdentifier] == null)
```
hillalex: This null check was carried out [here](https://github.com/vimc/montagu-api/pull/418/files#diff-7ad6d3cde953d76738a50c8a0f770406L199) but it could have changed between then and now. Effectively this just make the race condition tighter and manual testing indicates that it does work so is a good intermediate solution. A really robust solution will be to make an initial endpoint call that just initialises the cache before any file chunks are posted, which will require some front-end work as well.

weshinsley: What about making that section in BurdenEstimateUploadController.kt from lines 206-219 a critical section? (So only one thread at a time can ask if cachedMetaData == null - and the second asker is blocked until the first one finishes that section?

weshinsley: Eg (slightly coarsely), BurdenEstimateUploadController.kt :180
private fun getFileMetadata(): ChunkedFile
->
private synhronized getFileMetadata(): ChunkedFile

src/app/src/main/kotlin/org/vaccineimpact/api/app/ChunkedFileCache.kt
```diff
@@ -29,7 +30,14 @@ class ChunkedFileCache(private val flushInterval: Long = TimeUnit.HOURS.toMillis
...
+        {
+            memoryCache[item.uniqueIdentifier] = MutablePair(item, System.currentTimeMillis())
+        }
+        else
+        {
+            memoryCache[item.uniqueIdentifier]!!.left.uploadedChunks.putAll(item.uploadedChunks)
```
EmmaLRussell: Does this new step correspond to something which would have happened if the memory cache entry hadn't initially evaluated to null (but since it was, has to be repeated here)?

When this method is called by getFileMetadata, having evaluated that the cache entry does not yet exist, it returns the item ChunkedFile which was passed to this function (and which, under the race condition, does not end up in the cache as its equivalent is already there). The caller of getFileMetada, uploadBurdenEstimateFile, sets uploadedChunks in this returned metadata. Is that OK under the race condition because of the extra step taken here?

src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/BurdenEstimates/BurdenEstimateUploadController.kt
```diff
@@ -78,6 +79,10 @@ class BurdenEstimateUploadController(context: ActionContext,
 
         val metadata = getFileMetadata()
 
+        val alreadyUploadedKeys = metadata.uploadedChunks.keys().toList().joinToString(",")
```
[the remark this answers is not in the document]
hillalex: Was going to leave it in for now, just in case it proves useful.

src/app/src/main/kotlin/org/vaccineimpact/api/app/ChunkedFileCache.kt
hillalex: Huh, well spotted. Suggests that this wasn't in fact a fix, and the working behaviour we saw on UAT was just a coincidence!

src/app/src/main/kotlin/org/vaccineimpact/api/app/ChunkedFileCache.kt
```diff
@@ -34,10 +34,6 @@ class ChunkedFileCache(private val flushInterval: Long = TimeUnit.HOURS.toMillis
         {
             memoryCache[item.uniqueIdentifier] = MutablePair(item, System.currentTimeMillis())
         }
-        else
```
hillalex: @EmmaLRussell you alerted me to this being completely obsolete - the uploadedChunks map hasn't been written to yet at this point.

src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/BurdenEstimates/BurdenEstimateUploadController.kt
```diff
@@ -215,7 +215,7 @@ class BurdenEstimateUploadController(context: ActionContext,
         else
         {
             chunkedFileCache.put(providedMetadata)
-            providedMetadata
+            chunkedFileCache[uniqueIdentifier]!!
```
hillalex: This should ensure the correct object reference is passed back to the caller

src/app/src/main/kotlin/org/vaccineimpact/api/app/ChunkedFileCache.kt
hillalex: Yes I think that would work

hillalex:
Ok, how does this look to you both? This branch is deployed to UAT so would be good if we can do some manual testing. I've done a few uploads and haven't seen the bug yet.

PR #419, opened 2020-02-11
VIMC-3421 remove annex foreign data wrapper

hillalex:
This is the set of changes to make the api work with https://github.com/vimc/montagu-db/pull/139
1. updates db to the relevant branch
1. removes the call to add an annex user mapping when bringing up the databases
1. removes all references to stochastic estimates


richfitz:
There are still quite a lot of references to annex left in the api code after this (looking at a few files like `src/databaseInterface/src/main/java/org/vaccineimpact/api/db/annex/tables/BurdenEstimateStochastic.java` but also things like `src/blackboxTests/src/test/resources/config.properties`). Do we want to think about how much we're removing annex here and completely chop it out for now?

hillalex:
> There are still quite a lot of references to annex left in the api code after this (looking at a few files like `src/databaseInterface/src/main/java/org/vaccineimpact/api/db/annex/tables/BurdenEstimateStochastic.java` but also things like `src/blackboxTests/src/test/resources/config.properties`). Do we want to think about how much we're removing annex here and completely chop it out for now?

Thanks, I had intended to fully remove it so must have missed those.

hillalex:
Should all be gone now

PR #423, opened 2020-03-16
vimc-3567 bug with column ordering in some burden estimate downloads

hillalex:
Looks like these were just not being sorted, so it was pure chance if they were in step with the headings :woman_facepalming:


hillalex:
> I can see how this means that the column headers for outcomes should now match the ordering of the actual columns coming from the query, but I'm a bit confused about cohort size. It gets explicitly added as a column to return from the query in the repo, but it won't be in the list of headers given to the Flexible Data Table in the logic class. Shouldn't it be? Though it looks like that was also the case before, just that it was always the last column before..

It won't be in the list of headers because it exists as named a field on the BurdenEstimate class - in the database it's treated just as a burden_outcome (so should be ordered by id along with the others) but in the kotlin code it's treated specially, and so will always be serialised first in the resulting CSV. Does that make sense?

hillalex:
Here's the line where the `BurdenEstimate` gets `cohortSize` populated - https://github.com/vimc/montagu-api/pull/423/files#diff-579de71255a98c36389616c61a2d6ac0L249

PR #425, opened 2020-07-06
VIMC-4115 bug fix - update tests to reflect new database constraint

hillalex:
[This db change](https://github.com/vimc/montagu-db/commit/732ae2e021a6bab16c85195aa432cf1cd2588df5) was a breaking change and a bunch of tests are failing. This PR regenerates the db interface and updates the test helpers.


src/databaseTests/src/test/kotlin/org/vaccineimpact/api/databaseTests/tests/burdenEstimateRepository/BurdenEstimateRepositoryTests.kt
```diff
@@ -231,68 +231,6 @@ abstract class BurdenEstimateRepositoryTests : RepositoryTests<BurdenEstimateRep
             modelRunParameterSet = null
     )
 
-
```
hillalex, line 234: These unused methods must have been missed when deleting all stochastic related code. Now the table is gone they were failing to compile.

scripts/run-orderly-web-deps.sh
```diff
@@ -10,12 +10,13 @@ root=$(realpath $here/..)
 # create orderly db
 if [[ -d $root/src/demo ]]
 then
-  echo "Orderly demo folder already exists, not re-creating it."
```
hillalex, line 13: The build was also failing here but I don't really understand why - first it was complaining about missing the demo db artifact from the orderly build because all the TC data had been lost, so I just removed the dependency on that build (seeing as we're migrating away from TC anyway), assuming that it would hit this condition and regenerate the demo db. But it was hitting this condition and finding that `/src/demo` already existed and I still can't for the life of me see why/how. I've changed it to regenerate the demo db every time but I still don't understand why this wasn't working as is!

EmmaLRussell: Is it hitting this code twice in the course of the TC builds? Once in blackbox tests and once from ..something else?

hillalex: That's what I wondered but couldn't see how it was

PR #427, opened 2020-08-20
vimc-4020 proof of concept for celery-java

hillalex:
In case you want to pick this up tomorrow, here is a proof of concept for using the celery-java package to queue the diagnostic report task. For the test to pass you have to run the task-queue deps and worker from that repo, since I haven't added a script to bring that up here yet.


src/config/default.properties
```diff
@@ -1,3 +1,6 @@
+celery_broker=amqp://guest@localhost/%2F
```
hillalex, line 1: nb - do we want the celery broker to be locked down at all with users/permissions?

PR #431, opened 2020-09-16
vimc-4020 wiring up the task queue

hillalex:
This PR adds a client to talk to the task queue and a test that it can do so.

The actual code addition here is very very simple! https://github.com/vimc/montagu-api/pull/431/files#diff-7b27d6b50cdbadf6cd51e5d99ff67945

Everything else is just ensuring test dependencies get set up correctly.


r-ash (approved):
I'm not super familiar with how all this stuff fits together but looks good from what I can see!

PR #432, opened 2020-09-17
vimc-4020 integrate task queue into burden estimate upload

hillalex:
This PR actually queues the diagnostic report task when a burden estimate set is marked as closed.

Contains commits from https://github.com/vimc/montagu-api/pull/431

After that is merged, master should be merged into this branch and the base of the PR set back to master.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,8 +26,7 @@ before_script:
 script:
   - ./src/gradlew -p src :testLibrary :stopTestAPI
 # need to stop and re-start orderly-web on host network for blackbox tests against app running on metal
-# and stop and restart the database so it doesn't have the task queue user anymore
-  - ./src/gradlew -p src :stopDatabase :stopOrderlyWeb :startDatabase
+  - ./scripts/stop-orderly-web.sh host
   - ./scripts/start-orderly-web.sh host
   - nohup ./src/gradlew -p src :run > /dev/null 2>&1 &
   - ./src/gradlew -p src :blackboxTests:test
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,7 @@ before_script:
   - sudo mkdir -p /etc/montagu/api/ && sudo touch /etc/montagu/api/go_signal
   - ./src/gradlew -p src :startDatabase :startOrderlyWeb :startTaskQueue :startTestAPI
 script:
-  - ./src/gradlew -p src :testLibrary :stopTaskQueue :stopTestAPI
+  - ./src/gradlew -p src :testLibrary :stopTestAPI
 # need to stop and re-start orderly-web on host network for blackbox tests against app running on metal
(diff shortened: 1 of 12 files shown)


.travis.yml
```diff
@@ -24,7 +24,7 @@ before_script:
   - sudo mkdir -p /etc/montagu/api/ && sudo touch /etc/montagu/api/go_signal
   - ./src/gradlew -p src :startDatabase :startOrderlyWeb :startTaskQueue :startTestAPI
 script:
-  - ./src/gradlew -p src :testLibrary :stopTaskQueue :stopTestAPI
```
hillalex, line 27: we now need the task q to be running during blackbox tests, so don't stop it here

src/app/src/main/kotlin/org/vaccineimpact/api/app/clients/CeleryClient.kt
```diff
@@ -7,7 +7,11 @@ import org.vaccineimpact.api.db.Config
 typealias CeleryTaskResult = Map<String, Map<String, Boolean>>
 typealias CeleryTaskArguments = Array<out Any>
 
-class CeleryClient
+interface TaskQueueClient {
```
hillalex, line 10: need an interface so it can be mocked in unit tests

r-ash (approved):
Just for my own understanding. It looks like here `closeEstimateSetAndReturnMissingRowError` gets called from 3 places. The `BurdenEstimateSetUploadController` either on `populatieBurdenEstimateSetFromFile` or `populateBurdenEstimateSet` or from `BurdenEstimatesController` on `closeBurdenEstimateSet`. I'm guessing these are called on actions from the UI? What do user do to trigger those actions?

hillalex:
> Just for my own understanding. It looks like here `closeEstimateSetAndReturnMissingRowError` gets called from 3 places. The `BurdenEstimateSetUploadController` either on `populatieBurdenEstimateSetFromFile` or `populateBurdenEstimateSet` or from `BurdenEstimatesController` on `closeBurdenEstimateSet`. I'm guessing these are called on actions from the UI? What do user do to trigger those actions?

* `populateBurdenEstimateSet` used exclusively as an API endpoint (i.e. not from the portal) to populate estimates, with or without auto-closing the set as well. It reads estimates directly from the POST body. If not auto-closing the set, can upload multiple times to the same set, and then make a request to close the set:
*  `closeBurdenEstimateSet` also corresponds to one API endpoint and isn't used by the portal, but designed to be used in conjunction with the above endpoint, if uploading multiple times to the same set (e.g. uploading one country at a time)
* `populateBurdenEstimateSetFromFile` is the endpoint that the portal uses from the upload estimates page, where the user selects a file, this first gets POSTed to the server in chunks and then once successfully uploaded, this endpoint is called to ingest the estimates from the uploaded file

PR #433, opened 2020-10-06
vimc-4242 Add uploader email to parameters send to diagnostic report task

EmmaLRussell:
Went down a blind alley of attempting to get the uploader email from the ComonProfile in the context, but that just seems to ahve the user name - so adding user repo as a parameter to the controller to fetch email from there.

### DIFF
--- a/src/app/src/main/kotlin/org/vaccineimpact/api/app/clients/CeleryClient.kt
+++ b/src/app/src/main/kotlin/org/vaccineimpact/api/app/clients/CeleryClient.kt
@@ -7,11 +7,12 @@ import org.vaccineimpact.api.db.Config
 typealias CeleryTaskResult = Map<String, Map<String, Boolean>>
 typealias CeleryTaskArguments = Array<out Any>
 
-interface TaskQueueClient {
-    fun runDiagnosticReport(group: String, disease: String, touchstone: String): Any
+interface TaskQueueClient
+{
+    fun runDiagnosticReport(group: String, disease: String, touchstone: String, vararg additionalRecipients: String): Any
 }
 
-class CeleryClient: TaskQueueClient
+class CeleryClient : TaskQueueClient
 {
     private val broker = Config["celery.broker"]
     private val backend = Config["celery.backend"]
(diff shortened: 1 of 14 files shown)


src/app/src/main/kotlin/org/vaccineimpact/api/app/clients/CeleryClient.kt
```diff
@@ -7,11 +7,12 @@ import org.vaccineimpact.api.db.Config
 typealias CeleryTaskResult = Map<String, Map<String, Boolean>>
 typealias CeleryTaskArguments = Array<out Any>
 
-interface TaskQueueClient {
-    fun runDiagnosticReport(group: String, disease: String, touchstone: String): Any
+interface TaskQueueClient
+{
+    fun runDiagnosticReport(group: String, disease: String, touchstone: String, vararg additionalRecipients: String): Any
```
hillalex, line 12: why vararg here, will other users have their emails passed in the future?

hillalex (commented):
Looks like something is very wrong with coverage  - maybe run the Jacoco task locally to see what's going on there? Are the reports not being correctly generated, or are they failing to get sent to codecov?

src/app/src/main/kotlin/org/vaccineimpact/api/app/clients/CeleryClient.kt
EmmaLRussell: Probably not, I was just making it consistent with the Celery task. But can limit to 1 as that's how it will be in the API in practice

hillalex:
> Looks like something is very wrong with coverage - maybe run the Jacoco task locally to see what's going on there? Are the reports not being correctly generated, or are they failing to get sent to codecov?

Maybe just re-running the Travis build, could be something intermittent?

EmmaLRussell:
How bizarre :cry: 

> > Looks like something is very wrong with coverage - maybe run the Jacoco task locally to see what's going on there? Are the reports not being correctly generated, or are they failing to get sent to codecov?
> 
> Maybe just re-running the Travis build, could be something intermittent?

hillalex (approved):

PR #435, opened 2020-10-07
Remove temporary fix for Buildkite agents

hillalex:
This removes the temporary setting of the PATH env var in the build script, as this has now been added to all build agents.


hillalex:
Closing in favour of properly named branch

PR #437, opened 2020-10-12
vimc-4318 method to deserialise coverage from CSV

hillalex:
So here's the tiniest of first PRs - a model for the coverage rows, support for deserializing the GAVISupportLevel and ActivityType enums (also a convenient way of validating them), and a method that just reads in a CSV from the POST body and returns a sequence of rows.

### DIFF
--- a/src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/CoverageController.kt
+++ b/src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/CoverageController.kt
@@ -2,23 +2,32 @@ package org.vaccineimpact.api.app.controllers
 
 import org.vaccineimpact.api.app.app_start.Controller
 import org.vaccineimpact.api.app.context.ActionContext
+import org.vaccineimpact.api.app.context.RequestBodySource
 import org.vaccineimpact.api.app.controllers.helpers.ResponsibilityPath
 import org.vaccineimpact.api.app.logic.CoverageLogic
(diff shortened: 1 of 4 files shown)


src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/CoverageController.kt
```diff
@@ -2,24 +2,33 @@ package org.vaccineimpact.api.app.controllers
...
+        private val postDataHelper: PostDataHelper = PostDataHelper()
 ) : Controller(actionContext)
 {
     constructor(context: ActionContext, repositories: Repositories)
             : this(context, RepositoriesCoverageLogic(repositories))
 
+    fun getCoverageDataFromCSV(): Sequence<CoverageIngestionRow>
+    {
+        val source = RequestBodySource(context)
+        return postDataHelper.csvData(from = source)
```
EmmaLRussell, line 29: Oh that's clever, does it work out the class to deserialize to from the method's return type?

EmmaLRussell (approved):

EmmaLRussell:
Looks good, did you want to merge it along with the submodule change?

src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/CoverageController.kt
hillalex: I know, clever compiler right!

PR #439, opened 2020-11-02
mrc-4347 Handle special vaccine program values in vaccine column during coverage import

EmmaLRussell:
The ticket refers to 'HepB3' but from the script I think it should just be 'HepB'
https://github.com/vimc/montagu-imports/blob/4624ca785525ae20808bf8918cd0bbfce93376fa/2020-03-10_op17_201912gavi_1/R/fun_transform.R#L95

The script also appears to expect 'Pentavelent' but from some of the ticket threads it seemed that 'Penta' might be expected too. 
I've made the checks case insensitive. 

UPDATE: clarified a couple of things with Xiang: https://teams.microsoft.com/l/message/19:1ed98499f20b43f7a7d3b25fb00f4fd8@thread.tacv2/1604396049906?tenantId=2b897507-ee8c-4575-830b-4f8267c3d307&groupId=993c3b2e-8a31-496a-a485-eb466f411c31&parentMessageId=1604396049906&teamName=VIMC&channelName=Science&createdTime=1604396049906 
'Rubella' rather than 'RCV1', and 'HepB' rather than 'HepB3'

### DIFF
--- a/src/app/src/main/kotlin/org/vaccineimpact/api/app/logic/CoverageLogic.kt
+++ b/src/app/src/main/kotlin/org/vaccineimpact/api/app/logic/CoverageLogic.kt
@@ -50,32 +50,37 @@ class RepositoriesCoverageLogic(private val modellingGroupRepository: ModellingG
         val genders = touchstoneRepository.getGenders()
         val setDeterminants = mutableListOf<Pair<ActivityType, String>>()
         val setIds = mutableListOf<Int>()
-        val records = rows.map {
-            val set = Pair(it.activityType, it.vaccine)
-            var setIndex = setDeterminants.indexOf(set)
-            if (setIndex == -1)
-            {
-                // All coverage uploaded by GAVI should be put into a coverage set with
-                // gavi support level "WITH". Individual coverage rows may nevertheless have
-                // gaviSupport = false where e.g. countries are funding their own programs
-                val newId = touchstoneRepository.createCoverageSet(touchstoneVersionId,
-                        it.vaccine, it.activityType, GAVISupportLevel.WITH)
-                setIds.add(newId)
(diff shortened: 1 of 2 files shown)


src/app/src/main/kotlin/org/vaccineimpact/api/app/logic/CoverageLogic.kt
```diff
@@ -148,6 +153,18 @@ class RepositoriesCoverageLogic(private val modellingGroupRepository: ModellingG
         return SplitData(splitData.structuredMetadata, tableData)
     }
 
+    private fun mapVaccineForCoverageSet(vaccine: String) : List<String>
+    {
+        // GAVI may provide vaccine programs rather tha individual vaccines in some coverage rows
```
hillalex, line 158: Super pedantic but these aren't vaccine programs, they're just combination vaccines (programs are the dose schedules, like MR1 vs MR2)

hillalex (approved):
Perfect. Pendantic edit to comment but otherwise all good to go!

src/app/src/main/kotlin/org/vaccineimpact/api/app/logic/CoverageLogic.kt
EmmaLRussell: Oh ok - but the things we're mapping include MR1 and MR2, so do you mean those are vaccine programs but Pentavalent is just a 3-in-1 vaccine combo, so calling them vaccine combinations would be more general and cover both?

hillalex: `MR` is also a combo vaccine - measles and rubella in one, but it's delivered in 2 doses, hence 1 and 2.

PR #440, opened 2020-11-03
vimc-4329 validation for coverage rows

hillalex:
This PR adds validation to the coverage ingestion:
* nicely formatted error messages where a foreign key violation occurs (relevant for vaccines, countries)
* missing and duplicate rows reported for routine coverage
* unexpected years reported for either routine or campaign coverage


src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqTouchstoneRepository.kt
```diff
@@ -213,27 +213,17 @@ class JooqTouchstoneRepository(
...
-                this.touchstone = touchstoneVersionId
-                this.name = "$vaccine: $vaccine, $support, $activity"
-                this.vaccine = vaccine
-                this.gaviSupportLevel = support
-                this.activityType = activity
-            }
-            record.store()
-            return record.id
-        }
-        catch(e: DataAccessException)
```
hillalex, line 230: Now that we have a generic way of returning foreign key errors, don't need this custom logic

src/app/src/main/kotlin/org/vaccineimpact/api/app/logic/CoverageLogic.kt
```diff
@@ -82,13 +120,65 @@ class RepositoriesCoverageLogic(private val modellingGroupRepository: ModellingG
...
+            throw BadRequest("Unexpected year: ${row.year}")
+        }
+        if (row.activityType == ActivityType.ROUTINE)
+        {
+            val countryLookup = expectedRowLookup[vaccine] ?: generateCountryLookup(countries)
+            val year = row.year.toShort()
+            // note that although this country validation only happens for routine coverage rows,
+            // the foreign key constraint will catch invalid countries for campaign data
+            val yearLookup = countryLookup[row.country]
+                    ?: throw BadRequest("Unrecognised or unexpected country: ${row.country}")
```
hillalex, line 148: Katy did actually say that it didn't matter if they gave us a superset of countries (when we ship out the burden estimates we only use countries in the `burden_estimate_country_expectation` table anyway) but it would actually complicate the logic here to explicitly allow non-GAVI73 countries so in fact it won't accept any extra countries for routine data.

hillalex:
wrong branch name :facepalm:

PR #441, opened 2020-11-03
vimc-4349 validation for coverage rows

hillalex:
This PR adds validation to the coverage ingestion:

* nicely formatted error messages where a foreign key violation occurs (relevant for vaccines, countries)
* missing and duplicate rows reported for routine coverage
* unexpected years reported for either routine or campaign coverage


src/app/src/main/kotlin/org/vaccineimpact/api/app/logic/CoverageLogic.kt
```diff
@@ -82,13 +120,65 @@ class RepositoriesCoverageLogic(private val modellingGroupRepository: ModellingG
...
-    constructor(repositories: Repositories) : this(repositories.modellingGroup,
-            repositories.responsibilities,
-            repositories.touchstone,
-            repositories.scenario)
+    private fun validate(vaccine: String,
+                         row: CoverageIngestionRow,
+                         countries: List<String>,
+                         expectedRowLookup: CoverageRowLookup): CoverageRowLookup
+    {
+        if (!expectedGAVICoverageYears.contains(row.year))
```
EmmaLRussell, line 137: This is getting called for both routine and campaign isn't it? I thought the expected years only applied to routine (I'm not really sure what the rules are for campaigns but the ticket says "campaigns can be any number of years") Is the expectation that campaign years can only be subset of routine years (won't accept any projected campaigns after 2030).

src/app/src/main/kotlin/org/vaccineimpact/api/app/logic/CoverageLogic.kt
```diff
@@ -82,13 +120,65 @@ class RepositoriesCoverageLogic(private val modellingGroupRepository: ModellingG
...
+        {
+            throw BadRequest("Unexpected year: ${row.year}")
+        }
+        if (row.activityType == ActivityType.ROUTINE)
+        {
+            val countryLookup = expectedRowLookup[vaccine] ?: generateCountryLookup(countries)
+            val year = row.year.toShort()
+            // note that although this country validation only happens for routine coverage rows,
+            // the foreign key constraint will catch invalid countries for campaign data
```
EmmaLRussell, line 146: But that just checks it's a country in the table, could still be a non-GAVI country? Again, I'm not really sure what the rules are here, just want to clarify!

src/app/src/main/kotlin/org/vaccineimpact/api/app/logic/CoverageLogic.kt
```diff
@@ -82,13 +120,65 @@ class RepositoriesCoverageLogic(private val modellingGroupRepository: ModellingG
...
+                         countries: List<String>,
+                         expectedRowLookup: CoverageRowLookup): CoverageRowLookup
+    {
+        if (!expectedGAVICoverageYears.contains(row.year))
+        {
+            throw BadRequest("Unexpected year: ${row.year}")
+        }
+        if (row.activityType == ActivityType.ROUTINE)
+        {
+            val countryLookup = expectedRowLookup[vaccine] ?: generateCountryLookup(countries)
```
EmmaLRussell, line 143: So we ingest coverage for any vaccine(s) which already exist in the db. Is the expectation that GAVI will provide coverage for different vaccines with separate uploads or all in one go? If the latter, I suppose we should also be validating that there are no missing vaccines.

EmmaLRussell (commented):
This all looks great, just a couple of questions to clarify the requirements

src/app/src/main/kotlin/org/vaccineimpact/api/app/logic/CoverageLogic.kt
hillalex: Yeah, that was my understanding, but I will double check...

src/app/src/main/kotlin/org/vaccineimpact/api/app/logic/CoverageLogic.kt
hillalex: Correct. And this is a discrepancy between campaign and routine that I've left in for purely convenience reasons. Sorry I thought I had written a comment on the PR to this effect but it was on the PR that I closed!

Katy said that it didn't matter if they gave us a superset of countries (when we ship out the burden estimates we only use countries in the burden_estimate_country_expectation table anyway) but it would actually complicate the logic here to explicitly allow non-GAVI73 countries so in fact it won't accept any extra countries for routine data. For campaign data it will accept extra valid countries.

src/app/src/main/kotlin/org/vaccineimpact/api/app/logic/CoverageLogic.kt
hillalex: Nope, explicitly need to allow for a subset of vaccines to be uploaded at one time as data is sometimes not available at the same time. The page in the portal will provide feedback on which vaccines we're still waiting on though.

src/app/src/main/kotlin/org/vaccineimpact/api/app/logic/CoverageLogic.kt
hillalex: I agree the discrepancy between routine and campaign is slightly weird here though.

src/app/src/main/kotlin/org/vaccineimpact/api/app/logic/CoverageLogic.kt
hillalex: Ok this seems like a correct assumption.

hillalex:
Ok, thought it would be better to keep country checking rules consistent between campaign and routine, so have changed that.

EmmaLRussell (approved):

PR #443, opened 2020-11-09
vimc-4238 add utc and scenario to celery task

hillalex:
The build is error-ing because of the updates to the celery task. This just adds UTC time and scenario id to the task so that it passes. I'm not totally sure if scenario id is what's wanted, or scenario description, but this gets the build passing for now at least.


EmmaLRussell:
Me neither, I guess it should be human-readable / whatever they currently see in the contrib portal.

>I'm not totally sure if scenario id is what's wanted, or scenario description, but this gets the build passing for now at least.

EmmaLRussell (approved):

PR #445, opened 2020-11-10
vimc-4365 endpoint to get coverage upload template

hillalex:
Was going to just add this in the portal, but figured it might actually be more useful to have the API return it so that if the coverage format changes, that gets reflected automatically without having to manually keep the portal template up to date.

### DIFF
--- a/src/app/src/main/kotlin/org/vaccineimpact/api/app/app_start/Router.kt
+++ b/src/app/src/main/kotlin/org/vaccineimpact/api/app/app_start/Router.kt
@@ -59,14 +59,25 @@ class Router(val config: RouteConfig,
 
         logger.info("Mapping $fullUrl to ${endpoint.actionName} on ${endpoint.controller.simpleName}")
 
-        when (endpoint.method)
+        if (contentType.contains("json"))
         {
-            HttpMethod.get -> Spark.get(fullUrl, contentType, route, this::transform)
-            HttpMethod.post -> Spark.post(fullUrl, contentType, route, this::transform)
-            HttpMethod.put -> Spark.put(fullUrl, contentType, route, this::transform)
-            HttpMethod.patch -> Spark.patch(fullUrl, contentType, route, this::transform)
-            HttpMethod.delete -> Spark.delete(fullUrl, contentType, route, this::transform)
-            else -> throw UnsupportedValueException(endpoint.method)
+            when (endpoint.method)
+            {
+                HttpMethod.get -> Spark.get(fullUrl, contentType, route, this::transform)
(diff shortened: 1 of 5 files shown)


src/app/src/main/kotlin/org/vaccineimpact/api/app/app_start/Router.kt
```diff
@@ -59,14 +59,25 @@ class Router(val config: RouteConfig,
...
+                HttpMethod.post -> Spark.post(fullUrl, contentType, route, this::transform)
+                HttpMethod.put -> Spark.put(fullUrl, contentType, route, this::transform)
+                HttpMethod.patch -> Spark.patch(fullUrl, contentType, route, this::transform)
+                HttpMethod.delete -> Spark.delete(fullUrl, contentType, route, this::transform)
+                else -> throw UnsupportedValueException(endpoint.method)
+            }
+        }
+        else
+        {
+            when (endpoint.method)
```
hillalex, line 76: For all other methods returning csv data we've used the `StreamSerializable` construction which bypasses the transform and writes the content directly to the output stream. In this case that whole construction would be overkill, just want to return a raw string with no transformation, which apparently we haven't done before.

hillalex:
Of course the lines codecov are complaining about are hit during blackbox tests. We need to find a way of getting codecov properly configured so it recognises these tests!

src/app/src/test/kotlin/org/vaccineimpact/api/tests/controllers/coverage/PopulatingCoverageTests.kt
```diff
@@ -14,6 +14,13 @@ import org.vaccineimpact.api.test_helpers.MontaguTests
 
 class PopulatingCoverageTests : MontaguTests()
 {
+    @Test
+    fun `can get coverage upload template`() {
+        val sut = CoverageController(mock(), mock(), PostDataHelper())
+        val result = sut.getCoverageUploadTemplate()
+        assertThat(result).isEqualTo("\"vaccine\", \"country\", \"activity_type\", \"gavi_support\", \"year\", \"age_first\", \"age_last\", \"gender\", \"target\", \"coverage\"")
```
EmmaLRussell, line 21: Wasn't there an extra flag for 'national' that was to go in? (In addition to the free text field which we agreed would go on the upload metadata.) One for a separate ticket?

hillalex: Yes, there's already a ticket for this in the sprint.

EmmaLRussell (approved):

PR #444, opened 2020-11-09
vimc-4364 add metadata about coverage sets

hillalex:
Saves a metadata entry for every coverage upload. 
Contains commits from https://github.com/vimc/montagu-api/pull/442

I've added a separate ticket for retrieving the filename, to keep this changeset small. https://mrc-ide.myjetbrains.com/youtrack/issue/VIMC-4387

PR #446, opened 2020-11-10
vimc-4390 endpoint to retrieve metadata about most recently uploaded coverage sets

hillalex:
This PR adds an endpoint `/touchstones/:touchstone-version-id/coverage/meta/` which returns metadata about the most recently uploaded coverage sets for each vaccine that has data in the touchstone. This is to reflect back in the portal to track upload progress. 

It also adds documentation on this endpoint, the `GET coverage/template` and the `POST coverage` endpoints which were overlooked in previous PRs.

This webmodels PR should be merged in at the same time: https://github.com/vimc/montagu-api/pull/446

~~Contains commits from https://github.com/vimc/montagu-api/pull/444~~

### DIFF
--- a/src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqTouchstoneRepository.kt
+++ b/src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqTouchstoneRepository.kt
@@ -250,6 +250,7 @@ class JooqTouchstoneRepository(
                 COVERAGE_SET_UPLOAD_METADATA.UPLOADED_ON)
                 .fromJoinPath(COVERAGE_SET, COVERAGE_SET_UPLOAD_METADATA)
                 .where(COVERAGE_SET.TOUCHSTONE.eq(touchstoneVersionId))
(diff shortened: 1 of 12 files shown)


src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqTouchstoneRepository.kt
```diff
@@ -243,6 +243,20 @@ class JooqTouchstoneRepository(
         return record.id
     }
 
+    override fun getCoverageUploadMetadata(touchstoneVersionId: String): List<CoverageUploadMetadata>
+    {
+        return dsl.select(COVERAGE_SET.VACCINE,
```
EmmaLRussell, line 248: Would it be helpful for these to come out of the db in any particular order for display to user? Could by by recency of upload or maybe by vaccine would be most useful for user checking which vaccines have/haven't been uploaded?

hillalex: Yeah, maybe vaccine ordered might be most useful?

PR #449, opened 2020-11-11
vimc-4362 accept and persist subnational flag on coverage

hillalex:
associated webmodels PR: https://github.com/vimc/montagu-webmodels/pull/94/files

~~contains commits from https://github.com/vimc/montagu-api/pull/448 which should be merged first~~


richfitz (approved):
This looks reasonable to me from following it, no way I could have written it though :-/

PR #451, opened 2020-12-07
vimc-4434 Allow years 2000-2100 in coverage upload, mandatory years are 2021-2030

EmmaLRussell:
This ticket:
* takes out [current year + 1] as expected start year, replace with 2021
* allows upload of any number of coverage rows covering years 2000-2100 (mandatory years are still 2021-2030)

### DIFF
--- a/src/app/src/main/kotlin/org/vaccineimpact/api/app/logic/CoverageLogic.kt
+++ b/src/app/src/main/kotlin/org/vaccineimpact/api/app/logic/CoverageLogic.kt
@@ -48,7 +48,8 @@ class RepositoriesCoverageLogic(private val modellingGroupRepository: ModellingG
             repositories.scenario,
             repositories.expectations)
 
-    private val expectedGAVICoverageYears = LocalDate.now().plusYears(1).year..2030
+    private val mandatoryGAVICoverageYears = 2021..2030
+    private val allowedGAVICoverageYears = 2000..2100
 
     override fun getCoverageSetsForGroup(groupId: String, touchstoneVersionId: String, scenarioId: String):
             ScenarioTouchstoneAndCoverageSets
@@ -70,7 +71,7 @@ class RepositoriesCoverageLogic(private val modellingGroupRepository: ModellingG
         for (country in countries)
         {
             val yearMap = YearLookup()
-            for (year in expectedGAVICoverageYears)
+            for (year in mandatoryGAVICoverageYears)
             {
                 yearMap[year.toShort()] = false
             }
@@ -143,7 +144,7 @@ class RepositoriesCoverageLogic(private val modellingGroupRepository: ModellingG
                          countries: List<String>,
                          expectedRowLookup: CoverageRowLookup): CoverageRowLookup
     {
-        if (!expectedGAVICoverageYears.contains(row.year))
+        if (!allowedGAVICoverageYears.contains(row.year))
         {
             throw BadRequest("Unexpected year: ${row.year}")
         }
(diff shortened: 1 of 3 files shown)


src/app/src/main/kotlin/org/vaccineimpact/api/app/logic/CoverageLogic.kt
```diff
@@ -48,7 +48,8 @@ class RepositoriesCoverageLogic(private val modellingGroupRepository: ModellingG
             repositories.scenario,
             repositories.expectations)
 
-    private val expectedGAVICoverageYears = LocalDate.now().plusYears(1).year..2030
+    private val mandatoryGAVICoverageYears = 2021..2030
```
mwoodbri, line 51: Obviously this will make 2021 mandatory even when we're in 2022... presumably that's ok?

src/app/src/test/kotlin/org/vaccineimpact/api/tests/logic/CoverageLogic/SaveCoverageTests.kt
```diff
@@ -172,37 +175,42 @@ and 13 others"""
...
+
         val sut = RepositoriesCoverageLogic(mock(), mock(), mockRepo, mock(), mockExpectationsRepo)
-        val expectedMessage = "Unexpected year: 2010"
 
+        val testSequenceEarly = sequenceOf(
+                CoverageIngestionRow("HepB", "AFG", activityType, true, 1999, 1, 10, GenderEnum.BOTH, 100F, 65.5F, false)
+        )
+
+        val testSequenceLate = sequenceOf(
+                CoverageIngestionRow("HepB", "AFG", activityType, true, 21019, 1, 10, GenderEnum.BOTH, 100F, 65.5F, false)
```
mwoodbri, line 200: `21019` _looks_ a bit odd, given that the assertion on L209 is expecting `2101` - but I might be missing something?!

mwoodbri (commented):
Just a couple of queries

src/app/src/main/kotlin/org/vaccineimpact/api/app/logic/CoverageLogic.kt
EmmaLRussell: I think that's ok for now, it does mean we'll need to update it next year. No doubt other things will have changed by then!

src/app/src/test/kotlin/org/vaccineimpact/api/tests/logic/CoverageLogic/SaveCoverageTests.kt
EmmaLRussell: Oops! Yeah, you're right, that should have failed the test. I'll have look.

src/app/src/main/kotlin/org/vaccineimpact/api/app/logic/CoverageLogic.kt
mwoodbri: No problem - the alternative (i.e. current year to 2030) may not actually be better.

src/app/src/test/kotlin/org/vaccineimpact/api/tests/logic/CoverageLogic/SaveCoverageTests.kt
EmmaLRussell: Ha, this actually passed the test because it was only checking for a partial message since they whole thing is a bit long-winded - and "Unexpected year: 2101" is indeed a substring of "Unexpected year: 21019... etc" :facepalm:

mwoodbri (approved):

PR #452, opened 2020-12-08
vimc-4431 Be more strict about validation when deserializing Booleans

EmmaLRussell:
We were using the default java toBoolean method to deserialize booleans (for all our deserializations, not just coverage). This is actually very lenient, and interprets any string which is not equal to "true" (ignoring case) as a false value. 

I've introduced a validate method of our own for booleans, which throws an exception if the input is nether "true" nor "false" (ignoring case). 

This will impact coverage, as intended. I don't believe this will impact burden estimate uploads, our other main deserialization usage, since no values in the burden estimates are booleans.

### DIFF
--- a/src/app/src/test/kotlin/org/vaccineimpact/api/tests/controllers/coverage/PopulatingCoverageTests.kt
+++ b/src/app/src/test/kotlin/org/vaccineimpact/api/tests/controllers/coverage/PopulatingCoverageTests.kt
@@ -5,16 +5,22 @@ import com.nhaarman.mockito_kotlin.doReturn
 import com.nhaarman.mockito_kotlin.mock
 import org.assertj.core.api.Assertions.assertThat
 import org.assertj.core.api.Assertions.assertThatThrownBy
+import org.assertj.core.api.Condition
 import org.junit.Test
 import org.vaccineimpact.api.app.context.ActionContext
 import org.vaccineimpact.api.app.context.InMemoryRequestData
 import org.vaccineimpact.api.app.controllers.CoverageController
-import org.vaccineimpact.api.app.requests.MultipartDataMap
 import org.vaccineimpact.api.app.repositories.TouchstoneRepository
-import org.vaccineimpact.api.models.*
+import org.vaccineimpact.api.app.requests.MultipartDataMap
+import org.vaccineimpact.api.models.ActivityType
+import org.vaccineimpact.api.models.CoverageIngestionRow
+import org.vaccineimpact.api.models.CoverageUploadMetadata
+import org.vaccineimpact.api.models.GenderEnum
 import org.vaccineimpact.api.serialization.validation.ValidationException
 import org.vaccineimpact.api.test_helpers.MontaguTests
 import java.time.Instant
+import java.util.function.Predicate
+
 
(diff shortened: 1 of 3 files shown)


src/app/src/test/kotlin/org/vaccineimpact/api/tests/controllers/coverage/PopulatingCoverageTests.kt
```diff
@@ -74,6 +80,27 @@ class PopulatingCoverageTests : MontaguTests()
...
+            on { getParts(anyOrNull()) } doReturn MultipartDataMap(
+                    "description" to InMemoryRequestData("some description"),
+                    "file" to InMemoryRequestData(invalidGaviSupportCSVDataString)
+            )
+        }
+        val sut = CoverageController(mockContext, mock(), mock())
+        val exceptionCondition =  Condition<Throwable>(
+                Predicate<Throwable> { e: Throwable ->
+                    val ve = e as ValidationException
+                    ve.errors.count() == 1 && ve.errors[0].code == "csv-bad-data-type:1:gavi_support"
```
EmmaLRussell, line 96: DataTableDeserializer catches and rethrows the exception thrown by the new Deserializer code [here](https://github.com/vimc/montagu-api/blob/master/src/serialization/src/main/kotlin/org/vaccineimpact/api/serialization/DataTableDeserializer.kt#L99), so we expect the exception here to be slightly different from that thrown by the new code.

src/serialization/src/main/kotlin/org/vaccineimpact/api/serialization/Deserializer.kt
```diff
@@ -36,11 +38,22 @@ class Deserializer
...
+    private fun String.toBooleanValidate(): Boolean
+    {
+        if (this.equals("true", ignoreCase = true)) {
+            return true;
+        }
+        if (this.equals("false", ignoreCase = true)) {
+            return false;
+        }
+        throw ValidationException(listOf(ErrorInfo("invalid-boolean", "Unable to parse '$this' as Boolean")))
+    }
```
mwoodbri, line 58: Semicolons look a bit out of place here. But could perhaps go even further...

```suggestion
    private fun String.toBooleanValidate() = when
    {
        equals("true", ignoreCase = true) -> true
        equals("false", ignoreCase = true) -> false
        else -> throw Exception("Unable to parse '$this' as Boolean")
    }
```

mwoodbri (approved):
Looks ready to go! One suggestion relating to idiomatic Kotlin. Please use your discretion.

src/serialization/src/main/kotlin/org/vaccineimpact/api/serialization/Deserializer.kt
EmmaLRussell: Ugh, yeah, I always mindlessly pepper Kotlin with semicolons whenever I've been writing javascript. I think I'll see if I can get the linter to be more strict about them just to try and train myself out of that if nothing else. 

Ah, I did wonder about  using a `when`. Think I tend to under-use them as sometimes find them less readable than they could be, but you're right, that's much cleaner. Thanks!

PR #454, opened 2020-12-09
vimc-4465 Add activityType to CoverageUploadMetadata

EmmaLRussell:
Retrieve activityType names for coverage sets when returning metadata. Group by activity type as well as vaccine. Only the latest coverage set per vaccine and activity Type are returned.

Should be merged alongside the change to the models repo:
vimc/montagu-webmodels#95

This change is required to complete https://github.com/vimc/montagu-webapps/pull/444


mwoodbri (approved):
This all looks good. I was initially confused as to why there was no schema change necessary... until I realised that the activity type is per row, and that we aggregate in order to show latest. Ready to merge!

PR #455, opened 2021-01-28
vimc-4531 Negative value check

EmmaLRussell:
Disallow negative values for any outcome from being uploaded as part of burden estimates. A burden estimate set containing any negative values will be rejected with an exception. Can be tested manually by targeting this branch and running the montagu webapps locally. 

There's some ongoing discussion about whether null or 'NA' values should also be rejected, so I'm just going to keep this ticket to its original scope of checking for negatives. 0 values are allowed.

### DIFF
--- a/docs/spec/BurdenEstimates.md
+++ b/docs/spec/BurdenEstimates.md
@@ -103,6 +103,8 @@ based on which outcomes you wish to upload values for. More or fewer columns
 are allowed so long as all the outcome columns correspond to allowed burden
 outcomes in the database.
 
+Negative values are not allowed in outcome columns.
+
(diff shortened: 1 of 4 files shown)


mwoodbri (approved):
Works as expected - ready to merge!

For future reference, tested as follows:
~~~
cd montagu-webapps
echo vimc-4531 >config/api_version
app/scripts/run-development-dependencies.sh
npm run webpack --prefix app
npm run contrib --prefix app
~~~
Then upload [this file](https://github.com/vimc/montagu-api/files/5904070/vimc-4531.txt) as "Single model run" on [this page](http://localhost:5000/contribution/IC-Garske/responsibilities/op-2018-1/burdens/yf-campaign/)

PR #456, opened 2021-03-01
vimc-4617 Do not return closed expectations for closed responsibilities

EmmaLRussell:
Closing responsibilites within a responsibility set caused an error in the contribution portal when loading the burden estimates template page. 

This was because the Expectations respository was not checking `is_open` flag on responsibilities, and was returning applicable scenarios for those expectations belonging to closed reponsibilities. However only open responsibilities  were returned in the `responsibilities` part of the response so there was an internal inconsistency in the data which caused the front end code to fail. This branch adds this check of `is_open` so all returned data is consistent. 

To reproduce the bug locally:
- Run development dependencies for montagu-webapps (see instructions in that repo)
- Open the montagu db with: `docker exec -it montagu_db_1 psql -U vimc -d montagu`
- Close a responsibility with: `update responsibility set is_open = FALSE where id = 6;`  This will close the `yf-campaign` responsibility for touchstone `op-2018-1`.
- Run the contrib portal
- Browse to: http://localhost:5000/contribution/IC-Garske/responsibilities/op-2018-1/templates/
- You should see the page fail to load, with a console error: 
```
 Uncaught TypeError: Cannot read property 'id' of undefined
    at ExpectationsDescription.scenarioMapping (ExpectationsDescription.tsx:17)
    at ExpectationsDescription.tsx:45
    ...
```

To test that the fix in this branch works:
- Repeat all of the above process, but this time set the value in `config/api_version` to `vimc-4617` before running dependencies to load the image for this branch
- Browsing to http://localhost:5000/contribution/IC-Garske/responsibilities/op-2018-1/templates/ should load the page successfully, and a template should be shown for the `yf-routine` scenario only

### DIFF
--- a/scripts/orderly-web.yml
+++ b/scripts/orderly-web.yml
@@ -5,12 +5,14 @@ network: db_nw
 
 volumes:
   orderly: orderly_volume
+  redis: orderly_web_redis_data
 
 orderly:
   image:
     repo: vimc
     name: orderly.server
     tag: master
+    worker_name: orderly.server
   initial:
     source: clone
     url: https://github.com/vimc/orderly-demo
--- a/scripts/orderly-web.yml
+++ b/scripts/orderly-web.yml
@@ -7,6 +7,12 @@ volumes:
   orderly: orderly_volume
   redis: orderly_web_redis_data
 
+redis:
+  image:
+    name: redis
+    tag: "5.0"
+  volume: orderly_web_redis_data
+
 orderly:
   image:
     repo: vimc
--- a/src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqExpectationsRepository.kt
+++ b/src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqExpectationsRepository.kt
@@ -74,6 +74,7 @@ class JooqExpectationsRepository(dsl: DSLContext)
                 .where(
                         RESPONSIBILITY_SET.TOUCHSTONE.eq(touchstoneVersion)
                                 .and(RESPONSIBILITY_SET.MODELLING_GROUP.eq(modellingGroup))
+                                .and(RESPONSIBILITY.IS_OPEN)
                 )
                 .groupBy { it[Tables.expectations.ID] }
                 .map { getBasicDataAndMappingFromRecords(it.value) }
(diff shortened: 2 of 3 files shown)


mwoodbri (approved):
Sorry for the delay... quite a steep learning curve here. The best high-level reference I found for the relevant concepts seems to the docs for the [R library](https://www.vaccineimpact.org/montagu-r/articles/montagu_user_guide.html), combined with the [database schema](https://www.vaccineimpact.org/montagu-db-docs/latest/index.html).

Anyway... your explanation makes sense - thanks for the detail.

One specific question: what causes a responsibility to be closed, even if the touchstone is still open? Simply that the burden estimates have been provided/reviewed?

I guess the usual broader (rhetorical) questions apply:
- Is there anywhere else this fix needs to be made? (presumably not)
- Can we learn any lessons from the fact that we were running two (inconsistent) queries to compile a response? Presumably the other one was [here](https://github.com/vimc/montagu-api/blob/e942e482ed08b101c7810229f0e7dea5069d99d3/src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqResponsibilitiesRepository.kt#L204).

EmmaLRussell:
> One specific question: what causes a responsibility to be closed, even if the touchstone is still open? Simply that the burden estimates have been provided/reviewed?

This is done manually by Xiang (or Wes or Rob) setting the flag in the db, in this case to take out test responsibilities. I get the impression this has rarely been used in the past, hence this bug showing up now, and was perhaps something of a stopgap when first introduced. Responsibilities shouldn't really change within a touchstone, it just seems to be an easy way of removing them from a responsibility set. Or it would be if it worked!

> 
>     * Is there anywhere else this fix needs to be made? (presumably not)
> 
I don't *think* so. `getAllExpectations` already uses the flag, as does the ResponsibilitiesRepository. 

>     * Can we learn any lessons from the fact that we were running two (inconsistent) queries to compile a response? Presumably the other one was [here](https://github.com/vimc/montagu-api/blob/e942e482ed08b101c7810229f0e7dea5069d99d3/src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqResponsibilitiesRepository.kt#L204).
Maybe to consider duplications between repository classes more!

mwoodbri:
> > One specific question: what causes a responsibility to be closed, even if the touchstone is still open? Simply that the burden estimates have been provided/reviewed?
> 
> This is done manually by Xiang (or Wes or Rob) setting the flag in the db, in this case to take out test responsibilities. I get the impression this has rarely been used in the past, hence this bug showing up now, and was perhaps something of a stopgap when first introduced. Responsibilities shouldn't really change within a touchstone, it just seems to be an easy way of removing them from a responsibility set. Or it would be if it worked!

Ok, thanks. I wondering whether this was part of the standard workflow and if so whether there should be some UI to ensure that it's done in a consistent manner. But seems like this was an exception - even if the reason for doing it isn't entirely clear!

Please don't hesitate to merge.

PR #457, opened 2021-03-12
VIMC-4567 Migrate burden estimate upload feed to Teams

mwoodbri:


src/app/src/main/kotlin/org/vaccineimpact/api/app/logic/Notifier.kt
```diff
@@ -3,27 +3,46 @@ package org.vaccineimpact.api.app.logic
...
+                    "scenarioId" to scenarioId,
+                    "burdenEstimateSetStatus" to burdenEstimateSetStatus.name,
+                    "responsibilitySetComplete" to responsibilitySetComplete,
+                    "touchstone" to touchstone
+            ))
         }
         catch (e: Exception)
         {
-            logger.warn("There was a problem sending the Slack message: ${e.message}")
+            logger.warn("There was a problem sending the Flow message")
```
EmmaLRussell, line 45: Would it be useful to include the parameters here for diagnostics (previous version included whole message)?

EmmaLRussell (approved):
This looks good to me, nice to push the message wording etc into Flow. Tiny comment about logging but good to merge as is. 

This is more a query about the Flow setup than the code so should maybe be directing this at @r-ash .. Here's what those notifications looks like to me (when a full set-of-sets completes):
Screenshot from 2021-03-22 11-00-08
- Seems to be a broken image, is there anything we can do about that?
- Is there a robot user we can use to post the messages to avoid confusion? Suspect Rob might end up fielding queries on these just because it looks like he posted the message!

src/app/src/main/kotlin/org/vaccineimpact/api/app/logic/Notifier.kt
mwoodbri: Reinstated - thanks

r-ash:
> This looks good to me, nice to push the message wording etc into Flow. Tiny comment about logging but good to merge as is.
> 
> This is more a query about the Flow setup than the code so should maybe be directing this at @r-ash .. Here's what those notifications looks like to me (when a full set-of-sets completes):
> Screenshot from 2021-03-22 11-00-08
> 
>     * Seems to be a broken image, is there anything we can do about that?
> 
>     * Is there a robot user we can use to post the messages to avoid confusion? Suspect Rob might end up fielding queries on these just because it looks like he posted the message!

re image - this was working for me, funny it doesn't show for you? The "adaptive card" which gets posted just had a URL in it to some image I found online. I think this has since been removed though? It probably isn't really necessary if we've removed the notifications for incomplete uploads.

also re point 2 no way of doing this as a another user which is a bit annoying. We could do something else like instead of comment have an alternative message on the card that gets posted?

mwoodbri:
> * Seems to be a broken image, is there anything we can do about that?

Odd - it's no longer working for me either, but it was last week and not immediately obvious what the problem is. Possibly the fact it's embedded from Google Drive. But I've removed it because as Rob says it doesn't add much now.

> * Is there a robot user we can use to post the messages to avoid confusion? Suspect Rob might end up fielding queries on these just because it looks like he posted the message!

I'll have a quick look into tweaking the existing card, making it clear when the responsibility set is complete without having to post a reply as Rob.

EmmaLRussell:
Just another thought - where is the Teams integration documented, if anywhere? Maybe somewhere on KB to document how to tweak card if one needs to? Then could include a link to that in Montagu API README.

mwoodbri:
> Just another thought - where is the Teams integration documented, if anywhere? Maybe somewhere on KB to document how to tweak card if one needs to? Then could include a link to that in Montagu API README.

https://github.com/vimc/montagu-api/pull/457#issuecomment-804017491

mwoodbri:
> I'll have a quick look into tweaking the existing card, making it clear when the responsibility set is complete without having to post a reply as Rob.

Card now includes a footnote if responsibility set is completed by this burden estimate set

PR #459, opened 2021-07-09
VIMC-4767 Add comments to responsibility sets

mwoodbri:
Two main changes here:
- Update logic that returns comments for a touchstone version to include any comments added to responsibility sets
- Modify URL used to attach a comment to a responsibility in order to make space for a URL to annotate a responsibility set whilst avoiding ambiguous URLs. I think this is actually more logical as these endpoints don't directly relate to the responsibility (set), rather to comments, which are distinct entities. And the new URLs are self-describing in the sense that they they're scoped to a touchstone version, group and scenario (if applicable).
  - Previously:
    - `/touchstones/{touchstone-id}/responsibilities/{group-id}/{scenario-id}/comments/`
  - Now:
    - `/touchstones/{touchstone-id}/comments/{group-id}/{scenario-id}/`
    - `/touchstones/{touchstone-id}/comments/{group-id}/`


EmmaLRussell (commented):
LGTM. Good call on the url change - if it's not obvious what effect it will have just from the url, the spec explains it fully.

EmmaLRussell (approved):

PR #465, opened 2021-11-04
return all burden estimate sets as responsibility rows

hillalex:
Return rows for all burden estimate sets, not just the latest, in `GET /touchstones/:touchstone/responsibilities/csv/`

I would have liked to test this manually using UAT but I can't get the vpn working :( Will try again to get that sorted today but in the meantime please could you run this branch up on UAT and verify working as expected?


EmmaLRussell (commented):
Looks good to me - deployed to UAT and can now see multiple rows per scenario in csv, for each of the uploads. 

Does there need to be a PR for the webmodels change as well?

hillalex:
> Looks good to me - deployed to UAT and can now see multiple rows per scenario in csv, for each of the uploads.
> 
> Does there need to be a PR for the webmodels change as well?

Fab, thank you. https://github.com/vimc/montagu-webmodels/pull/101

EmmaLRussell (approved):

PR #467, opened 2022-03-24
vimc-6279 Remove gavi73 restriction on coverage countries

EmmaLRussell:
Allow coverage to be uploaded for all countries in the database, not just gavi73 countries. However, we continue to require gavi73 countries as a minimum for any coverage sets, and also require all mandatory years for coverage sets uploaded for any non-required countries. 

This means that we need to distinguish between `requiredCountries` and `allowedCountries` when doing validation in `CoverageLogic`, and also that we might need to inject some new year maps into the CountryLookup (which is initially generated for all required countries) if coverage include non-required countries. This should ensure that missing row errors should be raised if non-required country coverage is missing some years. 

This branch has been deployed to UAT, and can successfully ingest the OP19 coverage file with Syria data which failed on Production (which can be found [here](https://montagu.vaccineimpact.org/reports/report/incoming-op19/20220324-111800-95e277dc/#downloads).


hillalex (approved):
This seems like a good solution!

PR #468, opened 2022-05-03
vimc-6441 upgrade gradle

hillalex:
Upgrade to gradle 7 and use a new compatible docker plugin

PR #469, opened 2022-05-04
vimc-6443 upgrade kotlin

hillalex:
Required some small changes due to deprecated syntax and some additional dependency declarations in the gradle files as seems like deps are no longer resolved transitively where the first one is a local project.

Also adds a test for the cli image

~~https://github.com/vimc/montagu-api/pull/468 should be merged first and the base here changed to master~~

### DIFF
--- a/buildkite/pipeline.yml
+++ b/buildkite/pipeline.yml
@@ -20,5 +20,8 @@ steps:
 
   - wait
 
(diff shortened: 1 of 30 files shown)


EmmaLRussell:
I think I'd better let you resolve the conflicts!

EmmaLRussell (approved):

PR #471, opened 2022-07-11
vimc-6329 let task queue clean up burden estimate files

hillalex:
* makes upload directory configurable (to use an absolute path in prod, local directory locally)
* still remove files from cache, i.e. the app forgets about them, but don't delete the files on disk

Once merged, should also merge https://github.com/vimc/montagu/pull/223 to configure the deployment


EmmaLRussell (approved):
LGTM

PR #474, opened 2022-10-04
vimc-6763 new activity type

hillalex:
To be merged after https://github.com/vimc/montagu-webmodels/pull/103 and alongside https://github.com/vimc/montagu-db/pull/169

### DIFF
--- a/docs/schemas/ActivityType.schema.json
+++ b/docs/schemas/ActivityType.schema.json
@@ -1,4 +1,4 @@
 {
 	"id": "ActivityType",
-	"enum": [ "none", "routine", "campaign" ]
+	"enum": [ "none", "routine", "campaign", "routine-intensified", "campaign-reactive" ]
 }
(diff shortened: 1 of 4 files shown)


EmmaLRussell (approved):
I can't really remember how the database value 'routine-intensified' would get mapped to the new webmodels enum value ROUTINE_INTENSIFIED...? And I assume this doesn't have implications for the front end - I think in the download coverage page, we just render whatever activity type value is in the data?

hillalex:
> I can't really remember how the database value 'routine-intensified' would get mapped to the new webmodels enum value ROUTINE_INTENSIFIED...? And I assume this doesn't have implications for the front end - I think in the download coverage page, we just render whatever activity type value is in the data?

There's a mapper somewhere that maps the enum names to strings and vice versa (underscores to em dash and upper to lowercase). I added that db test case to confirm that each string does get mapped ok 9and that you can insert coverage with each of the expected values). And yeah, front-end doesn't reference this enum anywhere.

PR #477, opened 2023-01-19
vimc-6929 update ow api url

hillalex:
Updates OW and the task queue.

I also committed the gradle and compiler settings for Intellij, since opening up in a fresh install defaults to the worng version of java without these.

### DIFF
--- a/scripts/start-task-queue.sh
+++ b/scripts/start-task-queue.sh
@@ -18,12 +18,13 @@ docker run --rm -d \
   --name mq \
   redis
 
-docker pull vimc/task-queue-worker:vimc-6927
+TASK_QUEUE_WORKER=vimc/task-queue-worker:vimc-6927
+docker pull $TASK_QUEUE_WORKER
 docker run --rm -d \
   $NETWORK_MAPPING \
   -v $ROOT/scripts/task-queue-config.yml:/home/worker/config/config.yml \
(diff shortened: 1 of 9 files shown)


scripts/start-task-queue.sh
```diff
@@ -37,7 +38,7 @@ docker run -d \
 
 # add task q user
 CLI=vimc/montagu-cli:master
-docker pull $CLI
+#docker pull $CLI
```
EmmaLRussell, line 41: Should this be commented out?

src/config/blackboxTests/config.properties
```diff
@@ -1 +1 @@
-orderlyweb.api.url=http://orderly_web_web:8888/api/v1
+orderlyweb.api.url=http://orderly_web_web:8888/api/v2
```
EmmaLRussell, line 1: I guess this is going to need updating in montagu settings too?
https://github.com/vimc/montagu/tree/master/settings

hillalex: yes PR here https://github.com/vimc/montagu/pull/228

scripts/start-task-queue.sh
hillalex: no, my bad! will edit

EmmaLRussell (approved):

PR #481, opened 2025-07-29
mrc-6619 create packit users

EmmaLRussell:
This branch replaces Orderly Web with Packit in the one place where Montagu API was directly using OW, which was to create OW users when a Montagu user is created, so that an admin can immediately start assigning permissions and roles to the new user in Packit, rather than waiting for the first time that the user logs in (which would also trigger user creation). 

The core change is replacing `OrderlyWebAPIClient` with `PackitAPIClient`. I've pretty much just replaced like with like here, and have left in the eccentric polymorphous way of implementing certificate ignoring for localhost runs. 

A couple of other things to note:
- I did in the end use the secured endpoint to get the packit token rather than the unsecured one (which would have bypassed the proxy) as I needed to add the proxy to the dependencies anyway in order for the task queue tests to work, now that the task queue requires the proxy itself! That feels like the more "proper" way to do it anyway... (And that Montagu Monorepo looks ever more tempting...)
- I'm using venv when installing packit this time, though not on Buildkite as it didn't seem to like it. (Buildkite also seemed to need give the db container a longish interval to set up the db.)

This is deployed on uat along with the related [montagu-deploy](https://github.com/vimc/montagu-deploy/pull/17) and [montag-config](https://github.com/vimc/montagu-config/pull/20) branches. To test, you should be able to create a new user in the Montagu admin portal, then immediately see that user has been created in the Admin area of Packit.


.travis.yml
richfitz, line 1: goodness me.

buildkite/common
```diff
@@ -29,15 +29,14 @@ export DB_PORT=5432
...
-# Temporary measure to install constellation on all build steps - required by orderly-web install
-pip3 install constellation
-
 # Teardown on exit
 function cleanup() {
-  $HERE/../scripts/stop-database.sh
-  $HERE/../scripts/stop-api.sh
-  $HERE/../scripts/stop-orderly-web.sh
-  $HERE/../scripts/stop-task-queue.sh
+  $HERE/../scripts/stop-database.sh || true
```
richfitz: usually we do `set +e` here to disable errors and then you don't need the `|| true` bit

scripts/stop-proxy.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+set -ex
```
richfitz:
```suggestion
set -x
```

then you can drop the `|| true`

richfitz (commented):
This all looks reasonable, but I suggest getting someone like anmol to have a look through too? I ofc don't have admin rights on montagu so can't actually test this!

EmmaLRussell:
> This all looks reasonable, but I suggest getting someone like anmol to have a look through too? I ofc don't have admin rights on montagu so can't actually test this!

I've made you a user-manager on uat so you should be able to test now.. (Surprised admin didn't cover that though!)

richfitz:
thanks - that works

src/app/src/main/kotlin/org/vaccineimpact/api/app/clients/PackitAPIClient.kt
```diff
@@ -5,68 +5,74 @@ import com.google.gson.*
...
     }
 
-    private fun getOrderlyWebToken(): String
+    private fun getPackitToken(): String
     {
-        if (orderlyWebToken == null) {
-            val requestHeaders = mapOf(
-                    "Authorization" to "token $montaguToken",
-                    "Accept" to "application/json"
+        if (packitToken == null) {
```
absternator, line 56: bit confused of this check? when do we have a packitToken here?

absternator: there is also a class variable `packitToken` as well?... can we try avoid mirroring these names for confusion... also unsure the need for the class variable

absternator (commented):
Nice work looks good. just small comment. also couldnt test because i dont have permissions so maybe you can just show me it working :smile:

src/app/src/main/kotlin/org/vaccineimpact/api/app/clients/PackitAPIClient.kt
EmmaLRussell: Yeah, you're right, I think I just blindly copied this logic over from the OW implementation, but actually it should for sure be getting this token fresh each time so I'll remove the class variable.

EmmaLRussell: Actually, I've slightly changed my mind again. The client only lives for the duration of the action, so it shouldn't need to refresh the token once it has it. But it could just get the token as part of the constructor, rather than lazy-fetching it, then you wouldn't have that weird mirroring.

EmmaLRussell:
It didn't actually work to initialise in the constructor, apparently because of the polymorphism - `getHttpClient` returns null during `init` - even though I think that ought to work based on the kotlin docs. :shrug: 
Anyway, I've just replaced `getPackitToken` with `ensurePackitToken` so we can always use the class member rather than local variable. 

I've also stopped it from leaking error response details in responses out of the API

src/app/src/main/kotlin/org/vaccineimpact/api/app/clients/PackitAPIClient.kt
absternator: yeah that would be probably be good

absternator (approved):
LGTM as chatted

